PR #2, opened 2020-08-24
add nodegroupslist support

patrickshan:
* change NodeGroup object and cloud provider node group from 1:1 mapping to 1:N mapping which one NodeGroup object can have multiple cloud provider node groups


pkg/cloudprovider/aws/aws.go
```diff
@@ -24,14 +24,21 @@ const (
...
+func providerIDToInstanceID(providerID string) (instanceID string, err error) {
 	res := providerIDRegex.FindStringSubmatch(providerID)
 	if len(res) != 2 {
 		return "", fmt.Errorf("unable to extract instance ID from provider ID")
 	}
 	return res[1], nil
 }
 
+func instanceIDToProviderID(instanceID, availabilityZone string) (string, error) {
+	if len(instanceID) == 0 || len(availabilityZone) == 0 {
```
Jacobious52, line 36: `nit`: `instanceID == "" || availabilityZone == ""` more explicit in intent

pkg/cloudprovider/aws/aws.go
```diff
@@ -133,89 +142,165 @@ func (p *provider) TerminateInstance(providerID string) error {
...
+	instances := make(map[string]cloudprovider.Instance)
+	for _, group := range a.groups {
+		for _, i := range group.Instances {
+			providerID, err := instanceIDToProviderID(*i.InstanceId, *i.AvailabilityZone)
+			if err != nil {
+				a.logger.Info("[Instances] skip instance which failed instanceID to providerID conversion: %v", *i.InstanceId)
+				continue
+			}
+			instances[providerID] = &instance{
+				instance:      i,
```
Jacobious52, line 157: Will I get caught by the "for iterator" gotcha in Go? E.g. `i` always will end up pointing to the last element since it's not copied. Seems to be okay based on diff but just checking

pkg/cloudprovider/aws/aws.go
```diff
@@ -133,89 +142,165 @@ func (p *provider) TerminateInstance(providerID string) error {
...
-				outOfDate: a.instanceOutOfDate(i),
-			})
+// getInstanceNodeGroupByInstanceID finds the right autoscaling group for an instance based on
+// instance ID matches one instance ID from one of the node groups inside nodeGroupsList
+func (a *autoscalingGroups) getInstanceNodeGroupByInstanceID(instanceID string) (*autoscaling.Group, error) {
+	var targetNodeGroup *autoscaling.Group
+	foundNodeGroup := false
+	for _, group := range a.groups {
+		for _, i := range group.Instances {
+			if *i.InstanceId == instanceID {
```
Jacobious52, line 221: compare with `aws.StringValue(i.InstanceId)` instead

pkg/cloudprovider/aws/aws.go
```diff
@@ -133,89 +142,165 @@ func (p *provider) TerminateInstance(providerID string) error {
...
+	if foundNodeGroup {
+		return targetNodeGroup, nil
+	}
+	return nil, fmt.Errorf("failed to found target node group for instance with ID: %v", instanceID)
+}
+
+// getInstanceNodeGroupByGroupName finds the right autoscaling group for an instance based on
+// nodeGroup name passed in
+func (a *autoscalingGroups) getInstanceNodeGroupByGroupName(nodeGroup string) (*autoscaling.Group, error) {
+	if nodeGroup != "" {
```
Jacobious52, line 240: `nit` nesting and happy path leftmost

```
if nodeGroup != "" {
   return nil, fmt.Errorf("failed to found target node group for instance with group: %v", nodeGroup)
}

for ...
```

pkg/cloudprovider/aws/aws.go
```diff
@@ -133,89 +142,165 @@ func (p *provider) TerminateInstance(providerID string) error {
...
+		AutoScalingGroupName: group.AutoScalingGroupName,
 		InstanceIds:          aws.StringSlice([]string{instanceID}),
 	})
 
 	return verifyIfErrorOccured(apiErr, alreadyAttachedMessage)
 }
 
-func (a *autoscalingGroup) instanceOutOfDate(instance *autoscaling.Instance) bool {
+func (a *autoscalingGroups) instanceOutOfDate(instance *autoscaling.Instance) bool {
+	group, err := a.getInstanceNodeGroupByInstanceID(*instance.InstanceId)
```
Jacobious52, line 292: `aws.StringValue(instance.InstanceId)`

pkg/cloudprovider/aws/aws.go
```diff
@@ -248,8 +333,13 @@ func (i *instance) OutOfDate() bool {
 
 // MatchesProviderID returns if the instance ID matches the providerID
 func (i *instance) MatchesProviderID(providerID string) bool {
-	if instanceID, err := convertProviderID(providerID); err == nil {
+	if instanceID, err := providerIDToInstanceID(providerID); err == nil {
 		return *i.instance.InstanceId == instanceID
```
Jacobious52, line 337: aws.StringValue here too

pkg/controller/cyclenodestatus/transitioner/transitions.go
```diff
@@ -29,7 +29,7 @@ func (t *CycleNodeStatusTransitioner) transitionUndefined() (reconcile.Result, e
 	}
 	// This CRD should only be created by the controller, so we don't validate the method any further
 
-	// Check to ensure nodeGroupName is set
+	// Check to ensure NodeName is set
 	if len(t.cycleNodeStatus.Spec.NodeName) == 0 {
```
Jacobious52, line 33: not your code but `t.cycleNodeStatus.Spec.NodeName == ""` perhaps?

pkg/apis/atlassian/v1/common.go
```diff
@@ -9,3 +9,25 @@ func (in *CycleSettings) HasValidMethod() bool {
 		return false
 	}
 }
+
+// buildNodeGroupNames builds a union of cloud provider node group names
+// based on nodeGroupsList and nodeGroupName
+func buildNodeGroupNames(nodeGroupsList []string, nodeGroupName string) []string {
```
Jacobious52, line 15: Can we test this function?

docs/cycling/examples/cnr5.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: "atlassian.com/v1"
+kind: "CycleNodeRequest"
+metadata:
+  name: "example"
+  namespace: "kube-system"
+spec:
+  nodeGroupName: "az1-nodes.my-nodegroup.my-site"
```
Jacobious52, line 7: In this example what happens ? Is `nodeGroupName` the truth or `nodeGroupsList`? Should a CNR with both be invalid?

Jacobious52: Ah found it
> // buildNodeGroupNames builds a union of cloud provider node group names

This seems like reasonable backwards compatible behaviour 👍

Jacobious52 (changes requested):
lgtm just a few smaller changes and nits

pkg/cloudprovider/aws/aws.go
patrickshan: No. That problem was due to iterate a list of objects, copy the object into a temporary variable and append that temporary variable address into a list which is always the same. This one is a list of object pointer and the temporary variable `i` is also a pointer which will be updated for each iteration and that pointer will be added to that instance object.

pkg/cloudprovider/aws/aws.go
patrickshan: Actually that last error serves two different conditions: a) `nodeGroup == ""` b) no matching group found inside the existing groups list. 
But I can split that and add another error for the first condition.

pkg/apis/atlassian/v1/common.go
patrickshan: added tests for this function.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -78,24 +78,26 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
...
 	}
 
+	// get instances inside cloud provider node groups
+	nodeGroupInstances := nodeGroups.Instances()
 	// Do some sanity checking before we start filtering things
 	// Check the instance count of the node group matches the number of nodes found in Kubernetes
-	if len(kubeNodes) != len(nodeGroup.Instances()) {
+	if len(kubeNodes) != len(nodeGroupInstances) {
 		t.rm.LogEvent(t.cycleNodeRequest, "NodeCountMismatch",
 			"Instances in node group: %v, nodes in kube: %v",
```
awprice, line 97: We need to make this error clearer, as it may be hard to determine what node group is mismatched out of a list of node groups

awprice (changes requested):

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
```diff
@@ -0,0 +1,255 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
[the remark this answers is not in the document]
patrickshan: I re-run that script to generate these files as there are some new fields added. It seems these files have been renamed. wondering if it's a change from `operator-sdk` and I am using the same version mentioned inside go.mod:
```
❯ operator-sdk version
operator-sdk version: "v0.17.0", commit: "2fd7019f856cdb6f6618e2c3c80d15c3c79d1b6c", kubernetes version: "unknown", go version: "go1.13.10 darwin/amd64"
```

pkg/controller/cyclenoderequest/transitioner/transitions.go
patrickshan: can add more information related with this mismatch especially when cloud provider has more nodes than nodes in cluster (len(nodeGroupInstances) > len(kubeNodes)). If a node is detached from an ASG (len(kubeNodes)< len(nodeGroupInstances)), then there will be no information about the node group for that node. But I can add the node instance ID/provider ID in there.

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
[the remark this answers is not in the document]
Jacobious52: Yeah I think it's from the operator-sdk change. I just kept renaming the files manually back to the old ones 🤣

pkg/controller/cyclenoderequest/transitioner/transitions.go
patrickshan: added some extra information about nodes mismatch inside the log.

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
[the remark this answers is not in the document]
patrickshan: I think we should keep it updated with upstream. The good thing is that our document only mentions its parent directory when applying those CRDs which shouldn't be impacted by updating those file names.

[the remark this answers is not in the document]
awprice: @patrickshan I'm not concerned with the file name changing, it's the `apiVersion` that has changed to an older, deprecated version that is concerning.

awprice (changes requested):

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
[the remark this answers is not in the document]
patrickshan: @awprice I've tried to build `operator-sdk` locally with its `v0.17.x` branch which generates the same CRDs as the one i download from upstream:
```
❯ operator-sdk version
operator-sdk version: "v0.17.0", commit: "2fd7019f856cdb6f6618e2c3c80d15c3c79d1b6c", kubernetes version: "v1.17.2", go version: "go1.15 darwin/amd64"
```
and both of them build CRDs with apiVersion `v1beta1`.

Then I tried with v0.18.0 which generates `v1` apiVersion CRDs. Also talked with @vincentportella who has the same issue and change to v0.18.0 which generates the right CRDs.

I just pushed the ones generated by v0.18.0 client. I think going forward we should select one version which could consistently generate right CRDs.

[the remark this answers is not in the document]
patrickshan: Actually after reading https://sdk.operatorframework.io/docs/upgrading-sdk-version/v0.18.0/, it seems that `v1` is changed to be default in v0.18.0 but with v0.17.0 you can still specify that `v1` apiVersion using this:
```
operator-sdk generate crds --crd-version v1
```
Going to update that Makefile to use this. But after comparing the CRDs genereate by v0.18.0 and by 0.17.0 with that crd version parameter, they are the same.

awprice (approved):
LGTM!

Jacobious52 (approved):
lgtm

PR #4, opened 2020-09-01
Add Slack integration

vincentportella:
- Adding ability for Cyclops to post cycling updates to a Slack channel.


cmd/manager/main.go
```diff
@@ -31,9 +32,11 @@ var (
 
 	debug = app.Flag("debug", "Run with debug logging").Short('d').Bool()
 
-	cloudProviderName = app.Flag("cloud-provider", "Which cloud provider to use, options: [aws]").Default("aws").String()
-	addr              = app.Flag("address", "Address to listen on for /metrics").Default(":8080").String()
-	namespace         = app.Flag("namespace", "Namespace to watch for cycle request objects").Default("kube-system").String()
+	cloudProviderName     = app.Flag("cloud-provider", "Which cloud provider to use, options: [aws]").Default("aws").String()
+	messagingProviderName = app.Flag("messaging-provider", "Which message provider to use, options: [slack]").Default("slack").String()
```
Jacobious52, line 36: I feel like it would more clear for default to be `""` (empty/none)?
Then below don't build the provider if it's not set. Hiding the "not enabled" logic in the `BuildNotifier` return is a little obfuscated (nil, nil). Thoughts @awprice @patrickshan ?
```
if *messagingProviderName != "" {
    notifier, err := notifierbuilder.BuildNotifier(*messagingProviderName)
    if err != nil {
        log.Error(err, "Unable to build notifier")
	os.Exit(1)
    }
}
```

e.g. I had to dig into the Slack Builder code to see that 
```
// Return nil without an error if both tokens are missing since this is an optional feature
	if token == "" {
		return nil, nil
	}
```
Handling the on/off logic in the main will also be applicable across all providers, not just slack with a `TOKEN`

Also would make the help message make more sense if messaging is optional
```diff
-     --messaging-provider="slack"
+     --messaging-provider=""  
                                 Which message provider to use, options: [slack]
```

Jacobious52 (changes requested):
lgtm, just those two suggestions. Happy to discuss

cmd/manager/main.go
patrickshan: +1 for make it default to "" which makes it clearer for switching on/off notification feature.

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 0.0.3
+VERSION = 0.0.4
```
patrickshan, line 1: I was thinking about this version/tag when working on the last PR. Are we going to continue using this version/tag in github ? As we might have some overlap with our existing versions for cyclops in the future which might be confusing. Suggest we choose a different tag pattern, like `vX.Y.Z' here and might also want to bump this version a bit more since it has been running and tested internally for a while.

awprice: @Jacobious52 and I agreed that once internal changes have been pushed upstream and we have deprecated the internal copy of the repo we would release a "1.0.0" version.

Changing the the version pattern is out of the scope for this PR

cmd/manager/main.go
vincentportella: Agree on your points, handling the logic in the main is clearer. Also agree on making the default "".

pkg/controller/cyclenoderequest/controller.go
```diff
@@ -93,12 +99,32 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		if cycleNodeRequest.Status.SelectedNodes == nil {
+			cycleNodeRequest.Status.SelectedNodes = map[string]bool{}
+		}
+
+		// Extract the cluster name and add it to the cycleNodeRequest
+		if cycleNodeRequest.ClusterName == "" {
+			clusterName := os.Getenv(clusterNameEnv)
+
+			if clusterName == "" {
+				log.Error(err, "Missing cluster name")
```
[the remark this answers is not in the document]
vincentportella: Agreed, that an oversight on my part, I'll change it to return an error.

Makefile
patrickshan: Sounds good. Just need to be careful when we release 1.0.0 as it duplicates old image tag.

pkg/controller/cyclenoderequest/transitioner/node.go
```diff
@@ -85,6 +99,14 @@ func (t *CycleNodeRequestTransitioner) addNamedNodesToTerminate(kubeNodes []core
 		for _, kubeNode := range kubeNodes {
 			if kubeNode.Name == namedNode {
 				foundNode = true
+
+				t.cycleNodeRequest.Status.NodesAvailable = append(
+					t.cycleNodeRequest.Status.NodesAvailable,
+					v1.CycleNodeRequestNode{
```
patrickshan, line 105: There is an update for this object which adds a new `NodeGroupName` field : https://github.com/atlassian-labs/cyclops/blob/master/pkg/apis/atlassian/v1/cyclenoderequest_types.go#L73. 

That field will be used when a node is being attached back during healing process as now we support cycling multiple nodes groups and it needs to know which node group that node will be attached back to. 

That information can be found inside the nodeGroupInstances. similar case can be found here: https://github.com/atlassian-labs/cyclops/blob/master/pkg/controller/cyclenoderequest/transitioner/transitions.go#L145

patrickshan: Actually I missed that one below in my previous PR which also needs to be fixed.

vincentportella: I've added them

patrickshan (approved):
LGTM

awprice (approved):

Jacobious52 (approved):
lgtm. Thanks for those changes

PR #11, opened 2020-09-21
Fix 4 minor notifier related issues/bugs

vincentportella:
Fixing 4 minor issues/bugs

- Switched from using `cnr.Spec.NodeNames` to `cnr.Status.NodesToTerminate` in the progress field to account for `NodeNames` being empty when cycling an entire nodegroup. The progress field is left empty until the `NodesToTerminate` can be counted.
- Improved the progress field to count nodes that have been successfully cycled instead of counting the nodes that have been selected for cycling. A new field in the cnr is added: `cnr.Stats.NumNodesCycled`.
- Fixed the issue where `cnr.Spec.NodeGroupName` is empty because `cnr.Spec.NodeGroupList` is used instead. The nodegroup field now join all nodegroups listed across both cnr fields. As a result of this, the ordering of the field in the notification message has been changed to place the nodegroup field at the bottom.
- Fixed bug where `cnr.Status.NodesAvailable` isn't populated when an entire Nodegroup is cycled by leaving `cnr.Spec.NodeNames` empty. This oversight would allow cyclops to flip-flop between phases rapidly and cause it to hit slack API limits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION = 1.0.0
+VERSION = 1.0.1
 IMAGE = cyclops:$(VERSION)
 
 MANAGER_BIN = cyclops
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION = 1.0.1
+VERSION = 1.0.0
 IMAGE = cyclops:$(VERSION)
 
 MANAGER_BIN = cyclops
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 1.0.0
+VERSION = 1.0.1
```
awprice, line 1: I think we should only be bumping the version when cutting a release from master.

awprice (changes requested):

Jacobious52 (approved):

awprice (approved):

patrickshan (approved):

PR #12, opened 2020-09-22
Cut v1.0.1 release

vincentportella:
Update version to 1.0.1
Then create v1.0.1 tag after merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION = 1.0.0
+VERSION = 1.0.1
 IMAGE = cyclops:$(VERSION)
 
 MANAGER_BIN = cyclops


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 1.0.0
+VERSION = 1.0.1
```
awprice, line 1: I'd be inclined to say this should be a minor release - mainly due to the addition of the new field in the CRD. A patch/bug release ideally wouldn't introduce a new field.

awprice (changes requested):

Makefile
vincentportella: I'll push a new branch to name it correctly and close out this PR.

PR #18, opened 2020-10-20
Fixed flip-flopping when pod cannot be drained + added CNR name validation

vincentportella:
- Fixes flip-flopping behaviour when a `CNS` object times out due to a pod not getting drained. I added an explicit check for failed because the other option would have been to remove the label added by cyclops to the node, which I think should remain when it fails, as with other failure cases. Additionally, since `reapChildren` is indirectly called by `transitionFailed`, there should be not possibility of transitioning back to `Healing`.
- Fixes the transition to `Failed` loop by checking if the current and desired phases are the same when transitioning to Failed. If it is the case, it stops requeueing.
- Adds a check when creating a `CNR` object to make sure that the name is no longer than 63 characters to prevent creating/listing `CNS` objects with it as a label, which kubernetes explicitly does not support.
- Fixes small visual bug with the `Nodegroups` section in the slack status message. Fixes another to remove the stylised code section when no final error message is exists.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -40,6 +40,12 @@ func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result, 
 		return t.transitionToHealing(fmt.Errorf("selector cannot be empty"))
 	}
 
+	// Protect against failure case where cyclops checks for leftover CycleNodeStatus objects using the CycleNodeRequest name in the label selector
+	// Label values must be no more than 63 characters long
+	if len(t.cycleNodeRequest.Name) > 63 {
+		return t.transitionToFailed(fmt.Errorf("cycleNodeRequest name must be no more than 63 characters"))
+	}
+
 	// Transition the object to pending
 	return t.transitionObject(v1.CycleNodeRequestPending)
 }
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -26,18 +26,18 @@ var (
 func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result, error) {
 	t.rm.LogEvent(t.cycleNodeRequest, "Initialising", "Initialising cycleNodeRequest")
 
+	if t.rm.Notifier != nil {
+		if err := t.rm.Notifier.CyclingStarted(t.cycleNodeRequest); err != nil {
+			t.rm.Logger.Error(err, "Unable to post message to messaging provider", "phase", t.cycleNodeRequest.Status.Phase)
+		}
+	}
+
 	// Check fields on the cycleNodeRequest for validity. We rely on the CRD validation rules
 	// to do most of the work here for us.
 
 	// Check to ensure a label selector has been provided
 	if t.cycleNodeRequest.Spec.Selector.Size() == 0 {
 		return t.transitionToHealing(fmt.Errorf("selector cannot be empty"))
-	}
-
-	if t.rm.Notifier != nil {
-		if err := t.rm.Notifier.CyclingStarted(t.cycleNodeRequest); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -34,10 +40,10 @@ func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result,
 		return t.transitionToHealing(fmt.Errorf("selector cannot be empty"))
 	}
 
-	if t.rm.Notifier != nil {
-		if err := t.rm.Notifier.CyclingStarted(t.cycleNodeRequest); err != nil {
-			t.rm.Logger.Error(err, "Unable to post message to messaging provider", "phase", t.cycleNodeRequest.Status.Phase)
-		}
+	// Protect against failure case where cyclops checks for leftover CycleNodeStatus objects using the CycleNodeRequest name in the label selector
+	// Label values must be no more than 63 characters long
+	if len(t.cycleNodeRequest.Name) > 63 {
```
Jacobious52, line 45: should use the kubernetes `validation` package. "k8s.io/apimachinery/pkg/util/validation"  validation.IsDNS1035Label probably for object names

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -136,7 +141,8 @@ func (t *CycleNodeRequestTransitioner) reapChildren() (v1.CycleNodeRequestPhase,
 	}
 
 	// If we've finished most of our children, go back to Initialised to add some more nodes
-	if t.cycleNodeRequest.Status.ActiveChildren <= t.cycleNodeRequest.Spec.CycleSettings.Concurrency/2 {
+	// If the next phase should be failed, skip this since transitioning back to initialised would be flip-flopping behaviour
+	if nextPhase != v1.CycleNodeRequestFailed && t.cycleNodeRequest.Status.ActiveChildren <= t.cycleNodeRequest.Spec.CycleSettings.Concurrency/2 {
```
Jacobious52, line 145: How does `t.cycleNodeRequest.Status.ActiveChildren <= t.cycleNodeRequest.Spec.CycleSettings.Concurrency/2` work? PArticular why divide concurrency by 2?

pkg/generation/cnr.go
```diff
@@ -25,6 +25,12 @@ func ListCNRs(c client.Client, options *client.ListOptions) (*atlassianv1.CycleN
 
 // ApplyCNR takes a cnr and optionally uses dry mode in the create request
 func ApplyCNR(c client.Client, drymode bool, cnr atlassianv1.CycleNodeRequest) error {
+	// Protect against failure case where cyclops checks for leftover CycleNodeStatus objects using the CycleNodeRequest name in the label selector
+	// Label values must be no more than 63 characters long
+	if len(cnr.Name) > 63 {
```
Jacobious52, line 30: should use the kubernetes validation package. "k8s.io/apimachinery/pkg/util/validation". As above

Jacobious52 (changes requested):

pkg/controller/cyclenoderequest/transitioner/util.go
vincentportella: If at least half of the node in the current cycle have finished cycling, then cyclops heads back to `Initialised` to pick up more. Problem here is that if there's a single node selected in the `CNR` and the draining times out, the `CNS` object gets deleted and `ActiveChildren` stays at 0 which will always trigger going back to `Initialised`.

awprice: @Jacobious52 `t.cycleNodeRequest.Status.ActiveChildren <= t.cycleNodeRequest.Spec.CycleSettings.Concurrency/2` allows us to pick up more work when the majority of nodes in the current batch of `concurrency` have been cycled.

If concurrency is 3, rather than waiting for the entire 3 to finish before starting more, we wait for the majority to finish, i.e. 2 before starting more cycling. 

It's a minor optimisation, but for larger concurrency numbers would make a difference in speed.

awprice (approved):

Jacobious52 (approved):

PR #21, opened 2020-10-30
Change CNR validation to check example name

vincentportella:
- Moved the name validation to the validation method, side effect is that there is no name validation for cyclops-observer, on the flip side, the cnr will immediately fail.
- Checking the name example which includes the suffix.


awprice (approved):
LGTM

Jacobious52 (approved):

PR #22, opened 2020-11-18
Observer to check cluster-autoscaler scaleUp events before creating new CNRs

cespo:
### Observer to check cluster-autoscaler scaleUp events before creating new CNRs

This PR fixes a race condition between cyclops and cluster-autoscaler where a new instance is started but is yet to join the cluster at the time the new CNR is created.

Adds two arguments that can be passed to the observer app:
- prometheus-address: address of a local or remote Prometheus instance
- node-startup-time: avg time elapsed between node startup and its availability in kube

The observer will consider safe to start a new CNR if last scaleUp event is older than <node-startup-time>
Ignores the check if Prometheus/cluster-autoscaler are not available in the cluster


cmd/observer/main.go
```diff
@@ -52,8 +54,10 @@ func newApp(rootCmd *cobra.Command) *app {
 		dryMode:           rootCmd.PersistentFlags().Bool("dry", false, "api-server drymode for applying CNRs"),
 		waitInterval:      rootCmd.PersistentFlags().Duration("wait-interval", 2*time.Minute, "duration to wait after detecting changes before creating CNR objects. The window for letting changes on nodegroups settle before starting rotation"),
 		checkInterval:     rootCmd.PersistentFlags().Duration("check-interval", 5*time.Minute, `duration interval to check for changes. e.g. run the loop every 5 minutes"`),
+		nodeStartupTime:   rootCmd.PersistentFlags().Duration("node-startup-time", 2*time.Minute, "duration to wait after a scaleUp event is detected"),
 		runImmediately:    rootCmd.PersistentFlags().Bool("now", false, "makes the check loop run straight away on program start rather than wait for the check interval to elapse"),
 		runOnce:           rootCmd.PersistentFlags().Bool("once", false, "run the check loop once then exit. also works with --now"),
+		prometheusAddress: rootCmd.PersistentFlags().String("prometheus-address", "prometheus", "Prometheus service address"),
```
awprice, line 60: I think this variable should be better named to indicate that it should be the prometheus address of cluster-autoscaler

cmd/observer/main.go
```diff
@@ -52,8 +54,10 @@ func newApp(rootCmd *cobra.Command) *app {
 		dryMode:           rootCmd.PersistentFlags().Bool("dry", false, "api-server drymode for applying CNRs"),
 		waitInterval:      rootCmd.PersistentFlags().Duration("wait-interval", 2*time.Minute, "duration to wait after detecting changes before creating CNR objects. The window for letting changes on nodegroups settle before starting rotation"),
 		checkInterval:     rootCmd.PersistentFlags().Duration("check-interval", 5*time.Minute, `duration interval to check for changes. e.g. run the loop every 5 minutes"`),
+		nodeStartupTime:   rootCmd.PersistentFlags().Duration("node-startup-time", 2*time.Minute, "duration to wait after a scaleUp event is detected"),
```
awprice, line 57: Description text should indicate it's after a cluster-autoscaler scale up event

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 1.1.1
+VERSION = 1.1.2
```
vincentportella, line 1: Keep the version as is, update the version in another PR specifically for cutting the new release after internal testing.

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	if err != nil {
+		// Prometheus might not be installed in the cluster. return true if it can't connect
+		klog.Errorln("Error creating client: %v\n", err)
+		return safeToStart
+	}
+
+	v1api := promv1.NewAPI(client)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	result, warnings, err := v1api.Query(ctx, "cluster_autoscaler_last_activity{activity='scaleUp'}", time.Now())
```
awprice, line 274: Does cluster-autoscaler provide last activities for specific node groups?

I can see an issue if node group `A` needs to be cycled, but can't be cycled because node group `B` keeps being scaled up.

Makefile
Jacobious52: 👍  A merge will create a new image under the git commit short hash. We can use it for internal soaking before releasing a new version.

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	if err != nil {
+		// cluster-autoscaler might not be installed in the cluster. return true if it can't find the metrics of run the query
+		klog.Errorln("Error querying Prometheus: ", err)
+		return safeToStart
+	}
+	if len(warnings) > 0 {
+		klog.Errorln("Warnings: ", warnings)
+	}
+
+	v := result.(model.Vector)
```
Jacobious52, line 284: if `result` is not a Vector this will panic and fail. Can catch this with 
```
if v, ok := result.(model.Vector); !ok {
   // log error and return
}
```

pkg/observer/controller.go
```diff
@@ -309,6 +370,15 @@ func (c *controller) Run() {
 	klog.V(3).Infof("waiting for %v to allow changes to settle", c.WaitInterval)
 	select {
 	case <-time.After(c.WaitInterval):
+		timeout := 0
```
awprice, line 373: These should be constants

pkg/observer/controller.go
```diff
@@ -309,6 +370,15 @@ func (c *controller) Run() {
...
+		retries := 20
+		for c.safeToStartCycle() == false {
```
awprice: This new logic of trying to apply the CNRs should be in a seperate function - as having another set of sleeping/waiting inside a select starts getting complicated.

pkg/observer/controller.go
```diff
@@ -309,6 +370,15 @@ func (c *controller) Run() {
...
+			time.Sleep(10 * time.Second)
+			timeout++
+			if timeout >= retries {
+				klog.Fatalf("Not safe to start CNR. %v retries exceeded", retries)
```
awprice: Does this crash the observer?

awprice (changes requested):

Makefile
cespo: Reverted back to 1.1.1

pkg/observer/controller.go
```diff
@@ -309,6 +370,15 @@ func (c *controller) Run() {
 	klog.V(3).Infof("waiting for %v to allow changes to settle", c.WaitInterval)
 	select {
 	case <-time.After(c.WaitInterval):
+		timeout := 0
+		retries := 20
+		for !c.safeToStartCycle() {
+			time.Sleep(10 * time.Second)
+			timeout++
+			if timeout >= retries {
+				klog.Fatalln("Not safe to start CNR. Retries exceeded", retries)
```
Jacobious52, line 379: Thoughts about logging Error and `return`ing thus skipping the create this loop vs `Fatal`? 

I tend to reserve Fatal for unrecoverable circumstances where continuing doesn't make sense, or could cause damage.

cmd/observer/main.go
cespo: I'm actually querying Prometheus for cluster-autoscaler metrics

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	// consider safe if there isn't any scaleUp event yet
+	if v.Len() == 0 {
+		klog.Infoln("No scaleUp events")
+		return safeToStart
+	}
+
+	scaleUpTime := v[v.Len()-1].Value.String()
+	t, err := stringToTime(scaleUpTime)
+	if err != nil {
+		klog.Fatalln("Error converting the time: ", err)
```
Jacobious52, line 294: Same comment about crashing the controller here

pkg/observer/controller.go
cespo: Unfortunately not. I looked into it and it only provides the timestamp of last event

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	}
+
+	lastScaleEvent := time.Since(t)
+	if lastScaleEvent <= c.NodeStartupTime {
+		safeToStart = false
+		klog.Infoln("Scale up event recently happened", c.NodeStartupTime)
+		return safeToStart
```
Jacobious52, line 301: no need to keep the state in a variable here 

`return false`

pkg/observer/controller.go
cespo: Ack

cmd/observer/main.go
cespo: Ack

pkg/observer/controller.go
cespo: Ok, I can pull it out

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	sec, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return time.Unix(sec, 0), nil
+}
+
+// query cluster-autoscaler metrics to figure out if it's safe to start a new CNR
+func (c *controller) safeToStartCycle() bool {
+	safeToStart := true
```
Jacobious52, line 261: suggest removing this variable and just using `return true` / `return false` directly for each branch. Easier to follow and make changes without tracking what the variable state is.

pkg/observer/controller.go
cespo: Yes. If the cluster is actively scaling up/down it might be safer to crash and wait a new cycle of the observer

pkg/observer/controller.go
cespo: ok, will add the check

pkg/observer/controller.go
cespo: 👍

pkg/observer/controller.go
cespo: 👍

pkg/observer/controller.go
```diff
@@ -309,6 +370,15 @@ func (c *controller) Run() {
 	klog.V(3).Infof("waiting for %v to allow changes to settle", c.WaitInterval)
 	select {
 	case <-time.After(c.WaitInterval):
+		timeout := 0
+		retries := 20
+		for !c.safeToStartCycle() {
+			time.Sleep(10 * time.Second)
```
Jacobious52, line 376: not a huge fan of time.Sleep on the main thread. This will block the context closing.

Suggest break out into own function and using a select to wait on the both the time and the context.
```
for ...
    select {
	case <-time.After(c.WaitInterval):
          // if timeout >= retries { ...
        case <-c.stopCh:
           // program wants to exi nowt. return control, don't cycle

Comment (vincentportella) [inline pkg/observer/controller.go:376]: Perhaps query using exponential backoff github.com/cenkalti/backoff.

File: pkg/observer/controller.go [hunk 14]
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	client, err := api.NewClient(api.Config{
+		Address: c.PrometheusAddress,
+	})
+	if err != nil {
+		// Prometheus might not be installed in the cluster. return true if it can't connect
+		klog.Errorln("Error creating client:", err)
+		return safeToStart
+	}
+
+	v1api := promv1.NewAPI(client)
```
Jacobious52, line 271: If you give the Controller struct a promAPI field, you can construct the client/api and do all this checking once at startup (in the main thread), instead of doing it each loop.

pkg/observer/controller.go
```diff
@@ -243,6 +247,63 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	result, warnings, err := v1api.Query(ctx, "cluster_autoscaler_last_activity{activity='scaleUp'}", time.Now())
+	if err != nil {
+		// cluster-autoscaler might not be installed in the cluster. return true if it can't find the metrics of run the query
+		klog.Errorln("Error querying Prometheus: ", err)
```
Jacobious52, line 277: `nit` Errorln/Warningln automatically add whitespace between arguments. Can remove your extra whitespace at the end

pkg/observer/controller.go
Jacobious52: Rather we'd use `client-go` provided backoff package instead
`https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/backoff.go`

pkg/observer/controller.go
Jacobious52: I'd prefer to wait for the new cycle of observer without the crash. Just change the control flow rather than crash the container. Thoughts?

pkg/observer/controller.go
```diff
@@ -276,6 +338,26 @@ func (c *controller) nextRunTime() time.Time {
...
+	err := backoff.Retry(func() error {
+		if !c.safeToStartCycle() {
+			klog.Error("Cluster autoscaler scaleUp event in progress. Retry...")
+			return errors.New("cluster-autoscaler event in progress")
+		}
+		return nil
+	}, b)
+
+	if err != nil {
+		klog.Errorln("there are stll cluster-autoscaler scaleUp events")
```
vincentportella, line 354: `still`*

vincentportella:
This might be pedantic but I think it would be better split the prom check into it's own package with an interface. At the moment you're requiring the cluster to run prometheus. While it's use may be widespread, we should follow the same the generic design as `cloud-provider`.

pkg/observer/controller.go
cespo: fixed

Jacobious52 (approved):
I'm happy with it as those first set of changes 👍 

I still would recommend remaining comments as nits / best practice, but I'll leave that up to you and the others to decide.

awprice (approved):
LGTM

vincentportella (approved):

PR #29, opened 2021-03-18
this commit fixes #28

### LINKED ISSUE #28: [BUG] cyclop-observer is using the wrong metrics for cluster-autoscaler autoscaling avoidance
**Describe the bug**
The current `cluster-autoscaler` metric `cluster-autoscaler_last_activity{activity='scaleUp'}` used by `cyclop-observer` to blocks any CNR creation does not provide the expected metric for the scaling up avoidance logic to work.

The metric is updated roughly every time the controller does a check if scaleUp is needed (which is every cycle) and the last_activity is also less than the default wait for pods startup time thus if activated, the cluster-autoscaling avoidance will always report a false positive and goes into a wait loop unless the `node-startup-time` flag is set to less than 20 seconds.

**To Reproduce**
Run the cyclop-observer connected to a prometheus server that provides cluster-autoscaler metric.
There is a nodegroup out of date

**Expected behavior**
The controller will gets stuck in a wait loop
```
I0318 15:36:08.402823   86972 controller.go:301] Scale up event recently happened
E0318 15:36:08.402859   86972 controller.go:347] Cluster autoscaler scaleUp event in progress. Retry...
I0318 15:36:11.895222   86972 controller.go:301] Scale up event recently happened
E0318 15:36:11.895260   86972 controller.go:347] Cluster autoscaler scaleUp event in progress. Retry...
I0318 15:36:14.784697   86972 controller.go:301] Scale up event recently happened
E0318 15:36:14.784733   86972 controller.go:347] Cluster autoscaler scaleUp event in progress. Retry...
```

dtnyn:
- Use `cluster_autoscaler_last_activity{activity='scaleDown'}` instead of `cluster_autoscaler_last_activity{activity='scaleUp'}`
- Change the scale up avoidance logic to detect a scale up event when the cluster-autoscaler no longer do any "check if scale down is needed" for more than 40 seconds (the average time gap per check loop is 30 seconds)
- Fixes a few comments to reflect the new change

### DIFF
--- a/pkg/observer/controller.go
+++ b/pkg/observer/controller.go
@@ -300,11 +300,11 @@ func (c *controller) safeToStartCycle() bool {
 	// cluster_autoscaler_last_activity values will update every ~30 seconds in non-scaling scenario
 	// set threshold detection time to 40 seconds to take account for any anomaly that may put a slightly longer time
 	lastScaleEvent := time.Since(t)
-	if lastScaleEvent > (time.Second * 40) {
+	if lastScaleEvent > time.Second*40 {
 		klog.Infoln("Scale up event recently happened")
 		return false
 	}
-	klog.Infoln("No scale up event")
+	klog.V(3).Infoln("No scale up event")
 
 	return true
 }
--- a/pkg/observer/controller.go
+++ b/pkg/observer/controller.go
@@ -272,7 +272,8 @@ func (c *controller) safeToStartCycle() bool {
 	v1api := promv1.NewAPI(client)
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	result, warnings, err := v1api.Query(ctx, "cluster_autoscaler_last_activity{activity='scaleUp'}", time.Now())
+	// scaleDown metric will stops updating and report the last time it's updated when the cluster is scaling up
+	result, warnings, err := v1api.Query(ctx, "cluster_autoscaler_last_activity{activity='scaleDown'}", time.Now())
 	if err != nil {
 		// cluster-autoscaler might not be installed in the cluster. return true if it can't find the metrics of run the query
 		klog.Errorln("Error querying Prometheus:", err)
@@ -283,9 +284,9 @@ func (c *controller) safeToStartCycle() bool {
 	}
 
 	v := result.(model.Vector)
-	// consider safe if there isn't any scaleUp event yet
+	// cluster-autoscaler should always gives a response if it's active
 	if v.Len() == 0 {
-		klog.Infoln("No scaleUp events")
+		klog.Errorln("Empty response from prometheus")
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/observer/controller.go
```diff
@@ -296,11 +297,14 @@ func (c *controller) safeToStartCycle() bool {
 		return false
 	}
 
+	// cluster_autoscaler_last_activity values will update every ~30 seconds in non-scaling scenario
+	// set threshold detection time to 40 seconds to take account for any anomaly that may put a slightly longer time
 	lastScaleEvent := time.Since(t)
-	if lastScaleEvent <= c.NodeStartupTime {
+	if lastScaleEvent > (time.Second * 40) {
```
Jacobious52, line 303: nit: don't need the brackets

pkg/observer/controller.go
```diff
@@ -296,11 +297,14 @@ func (c *controller) safeToStartCycle() bool {
...
 		klog.Infoln("Scale up event recently happened")
 		return false
 	}
+	klog.Infoln("No scale up event")
```
Jacobious52, line 307: use `klog.V(3).Infoln()`

Jacobious52 (changes requested):

Jacobious52 (approved):

pkg/observer/controller.go
```diff
@@ -296,11 +297,14 @@ func (c *controller) safeToStartCycle() bool {
 		return false
 	}
 
+	// cluster_autoscaler_last_activity values will update every ~30 seconds in non-scaling scenario
```
awprice, line 300: Where does the 30 seconds come from? Is this the same as the prometheus scrape_interval?

pkg/observer/controller.go
```diff
@@ -296,11 +297,14 @@ func (c *controller) safeToStartCycle() bool {
...
+	// set threshold detection time to 40 seconds to take account for any anomaly that may put a slightly longer time
 	lastScaleEvent := time.Since(t)
-	if lastScaleEvent <= c.NodeStartupTime {
+	if lastScaleEvent > time.Second*40 {
```
awprice, line 303: I think this duration should be in a var

pkg/observer/controller.go
```diff
@@ -272,7 +272,8 @@ func (c *controller) safeToStartCycle() bool {
 	v1api := promv1.NewAPI(client)
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	result, warnings, err := v1api.Query(ctx, "cluster_autoscaler_last_activity{activity='scaleUp'}", time.Now())
+	// scaleDown metric will stops updating and report the last time it's updated when the cluster is scaling up
```
awprice, line 275: I find this comment a little confusing - we are using the scaleDown dimension, but it is updated when the cluster is scaling up?

dtnyn: Ah yes, because last_activity scaleDown and scaleUp is updated when the controller is performing the activity of checking if it needs to scaleDown or scaleUp, not when it is scaling down and up. So the end effect is that both metric continuously gets updated every cycle if no event is happening. The scaleDown metric however does not get updated when the the cluster is scaling up since there is no need to perform any check for if scaleDown is needed when a scale up event is happening. 

The metric behavior is documented on the `cluster-autoscaler` repo https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/metrics.md#cluster-autoscaler-execution

pkg/observer/controller.go
dtnyn: Yes this is from the `scrape_interval` attribute

awprice: Yea `scrape_interval` could change depending on how a user has configured prometheus to scrape cluster-autoscaler - but we could have a CLI flag to change this, with a sane default of 30 or 40 seconds.

pkg/observer/controller.go
awprice: 👍 Thanks for the explanation!

awprice (approved):
LGTM!

Jacobious52 (approved):

PR #31, opened 2021-04-12
Add customisable timeout for cycling 

dtnyn:
Allows for user defined timeout for cycling timeout.

Changes:
* Added cyclingTimeout in cyclingSpec section of CRDs
* Controller accepts `default-cns-cycling-expiry` flag to set default timeout for cycling
* CLI now accepts `cycling-timeout` flag to set cycling timeout override for CNR generation
* fixed operator-sdk installation in Makefile


Makefile
```diff
@@ -52,10 +52,13 @@ lint:
 docker:
 	docker build -t $(IMAGE) .
 
+# New version of operator-sdk no longer support generate CRDs directly
+# Build from release v0.19.0 with commit hash 
 install-operator-sdk:
 	mkdir -p $(GOPATH)/src/github.com/operator-framework
 	-cd $(GOPATH)/src/github.com/operator-framework && git clone https://github.com/operator-framework/operator-sdk
 	git -C $(GOPATH)/src/github.com/operator-framework/operator-sdk checkout master
+	git -C $(GOPATH)/src/github.com/operator-framework/operator-sdk checkout 8e28aca60994c5cb1aec0251b85f0116cc4c9427
```
Jacobious52, line 61: able to pin to a tag or branch ? `v0.19.0` or something?

pkg/controller/cyclenodestatus/transitioner/util.go
```diff
@@ -39,5 +39,11 @@ func (t *CycleNodeStatusTransitioner) transitionObject(desiredPhase v1.CycleNode
 // timedOut returns true if the processing of this CycleNodeStatus has been going longer
 // than the timeout threshold
 func (t *CycleNodeStatusTransitioner) timedOut() bool {
-	return time.Now().After(t.cycleNodeStatus.Status.StartedTimestamp.Time.Add(nodeTerminationGracePeriod))
+	parsedCyclingTimeout, err := time.ParseDuration(t.cycleNodeStatus.Spec.CycleSettings.CyclingTimeout)
```
Jacobious52, line 42: can you parse out this time earlier such that you only have to call it once, and just store the Duration

Jacobious52 (changes requested):
looks good just those couple things.

Btw did you observe how this handles for really long timeout periods?

I'm wondering how noisey the logs will be after an order of weeks of waiting.

pkg/controller/cyclenodestatus/transitioner/util.go
dtnyn: Sounds good! Since this conversion would be called every re-queue

pkg/apis/atlassian/v1/common_types.go
```diff
@@ -37,4 +37,9 @@ type CycleSettings struct {
 	// IgnoreNamespaces is a list of namespace names in which running pods should be ignored
 	// when deciding whether a node has no more pods running.
 	IgnoreNamespaces []string `json:"ignoreNamespaces,omitempty"`
+
+	// CyclingTimeout is a string in time duration format that defines how long a until an
+	// in-progress CNS request timeout from the time it's worked on by the controller
```
awprice, line 42:
```suggestion
	// in-progress CNS request timeout from the time it's worked on by the controller.
```

awprice (changes requested):
LGTM, just one nit

awprice:
> Though there are some issues with approach, perhaps we could calculate re-queue time dynamically based on the cyclingTimeout we provided where we will also progressively decrease re-queue time once we're closer to the expected timeout date to make sure we don't miss any changes on the node when we finally drain.

An exponential back-off for generating how long the re-queue could work too - alot simpler. This should be a separate change however.

> Both of which will produce logs.

Probably worthwhile in a separate issue addressing the log verbosity for log lines that aren't needed in everyday operation.

Jacobious52:
I agree. I think it will be worthwhile addressing the verbosity with either or both those methods, in a separate Issue/PR.

I think however it is in scope of the underlying work ticket to make the solution viable before we roll out and use the change. So it might be worth following up in a second PR with verbosity changes.

dtnyn:
I realised since the apimachinary types package has built marshalling of time.duration to human readable format, I should make use of this instead of handling this ourselves with string parsing. I will make a quick change after validating it

Jacobious52 (approved):

awprice (approved):

PR #35, opened 2021-06-10
Force delete pods off the node if they have been unhealthy for a while when cycling

### LINKED ISSUE #34: [BUG] Deployment/ReplicaSet with 0 healthy replicas and PDB can cause stalling when draining nodes
**Describe the bug**
When Cyclops is draining a node, and a pod is crash looping on the node inside a deployment with a PDB of `maxUnavailable=1` while also having simultaneously unavailable replicas on other nodes, then that pod is unable to be removed and the node cycling will fail.

**To Reproduce**
Create a container in a deployment with 2 replicas that exits after a few seconds but will be restarted, and a PDB that looks like this:
```
NAMESPACE                         NAME                                          MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
namespace                    default--agcfp5le97a58h4e--agcfp5le97a58h4e   N/A             1                 1                     113m
```

If we try and cycle that node now, because both of the 2 replicas are unavailable the eviction API will refuse to evict either of the 2 replicas from the node despite the fact that these may never become healthy again.

**Expected behavior**
Because this pod is clearly already unhealthy for a large period of time, the controller should be smart enough to remove it and continue its job.

**Kubernetes Cluster Version**
v1.19.7

**Cyclops Version**
v1.15.0

mwhittington21:
Fixes #34

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION = 1.5.0
+VERSION = 1.6.0
 IMAGE = cyclops:$(VERSION)
 
 MANAGER_BIN = cyclops
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION = 1.6.0
+VERSION = 1.5.0
 IMAGE = cyclops:$(VERSION)
 
 MANAGER_BIN = cyclops
--- a/pkg/k8s/pod.go
+++ b/pkg/k8s/pod.go
@@ -2,6 +2,8 @@ package k8s
 
 import (
 	"context"
+	"time"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/policy/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 1.5.0
+VERSION = 1.6.0
```
awprice, line 1: We should bump the version and release independently after testing internally.

awprice (changes requested):
Logic looks good, really just some minor changes

pkg/k8s/pod.go
```diff
@@ -12,11 +14,27 @@ import (
...
+const (
+	podConditionTypeForUnhealthy = v1.PodReady
+)
+
 var log = logf.Log.WithName("k8s.pod.go")
 
-// EvictPod evicts a single pod from Kubernetes
+// KillAndRemovePod immediately terminates a given pod off the node, without waiting to check if it has been removed.
+// This should only be called on workloads that can tolerate being removed like this, and as a last resort.
+func KillAndRemovePod(podName, podNamespace, nodeName string, client kubernetes.Interface) error {
```
awprice, line 25: I think this should be `ForciblyDeletePod` - it better describes what we are actually performing on the pod.

Log lines below should be updated to match

pkg/k8s/pod.go
```diff
@@ -27,18 +45,39 @@ func EvictPod(pod *v1.Pod, apiVersion string, client kubernetes.Interface) error
 	})
 }
 
-// EvictPods evicts multiple pods from Kubernetes
-func EvictPods(pods []*v1.Pod, apiVersion string, client kubernetes.Interface) (evictionErrors []error) {
+// EvictOrKillPod tries to evict a pod, and if that fails will then check if it can forcibly remove the pod instead.
+func EvictOrKillPod(pod *v1.Pod, apiVersion string, client kubernetes.Interface, unhealthyAfter time.Duration, now time.Time) error {
```
awprice, line 49: nit: should be EvictOrForciblyDeletePod

awprice (approved):
LGTM

dtnyn (approved):

Jacobious52 (approved):

PR #37, opened 2021-10-20
Add cycling health checks

vincentportella:
- Added `healthChecks` to the `Nodegroup` and `CycleNodeRequest` specs to allow a user to specify custom health checks to perform during cycling.
- Health checks are performed at the beginning on cycling on all nodes selected to be terminated
- Health checks are performed during cycling before an old instance is terminate.
- Adding timeout cli argument to set the health check timeout period


docs/cycling/examples/README.md
```diff
@@ -201,4 +201,35 @@ spec:
...
+    validStatusCodes:
+    - 200
+    waitPeriod: 5m
+  - endpoint: http://service-name.namespace.svc.cluster.local:9090/ready
+    validStatusCodes:
+    - 200
+    waitPeriod: 5m
+```
+
+Cyclops can optionally perform a set of health checks before each node selected in terminated. This can be useful to perform deep health checks on system daemons or pods running on host network to ensure they are healthy before continuing with cycling. The set of health checks will be performed until each returns a healthy status once. `{{ .NodeIP }}` can be use to render the endpoint with the private ip of a new instance brought up during the cycling.
```
awprice, line 235:
```suggestion
Cyclops can optionally perform a set of health checks before each node selected is terminated. This can be useful to perform deep health checks on system daemons or pods running on host network to ensure they are healthy before continuing with cycling. The set of health checks will be performed until each returns a healthy status once. `{{ .NodeIP }}` can be use to render the endpoint with the private IP of a new instance brought up during the cycling.
```

pkg/apis/atlassian/v1/common_types.go
```diff
@@ -7,6 +7,9 @@ import (
 // CycleNodeRequestMethod is the method to use when cycling nodes.
 type CycleNodeRequestMethod string
 
+// HealthCheckScheme specifies the scheme used to connect to the node.
+type HealthCheckScheme string
```
awprice, line 11: Is this type used anywhere?

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -170,6 +170,60 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
 		}
 	}
 
+	if t.cycleNodeRequest.Spec.HealthChecks != nil {
```
awprice, line 173: I think this logic should be broken out into a discrete function

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -376,6 +431,96 @@ func (t *CycleNodeRequestTransitioner) transitionScalingUp() (reconcile.Result,
...
+		return t.transitionObject(v1.CycleNodeRequestCordoningNode)
+	}
+
+	var allHealthChecksPassed bool = true
+
+	// Find new instsances attached to the nodegroup and perform health checks on them
+	// before terminating the old ones they are replacing
+	// Cycling is paused until all health checks pass
+	// If no health checks are configured, nothing happens
+	for _, kubeNode := range kubeNodes {
```
awprice, line 446: Same again, should be broken out into a separate function

pkg/controller/cyclenoderequest/transitioner/transitions.go
vincentportella: Good point. The phases are getting a bit big

pkg/controller/cyclenoderequest/transitioner/healthchecks.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return renderedEndpoint.String(), nil
+}
+
+// healthCheckPassed checks if the statusCode returned matches the set of valid status code for the health check
+// as well as if the body matches regex provided
+func healthCheckPassed(healthCheck v1.HealthCheck, statusCode uint, body []byte) (bool, string) {
+	// If there is not regex match string specified, don't check the body of the response
+	r := regexp.MustCompile(healthCheck.RegexMatch)
```
Jacobious52, line 53: MustCompile panics if it fails. User input should not panic the controller, it should be handled gracefully and fail the CNR if user feedback. 

Use regexp.Compile which returns an error

pkg/controller/cyclenoderequest/transitioner/healthchecks.go
```diff
@@ -0,0 +1,222 @@
...
+	return uint(resp.StatusCode), bytes, nil
+}
+
+// performInitialHealthChecks on the nodes selected to be terminated before cycling begin. If any health
+// check fails return an error to prevent cycling from starting
+func (t *CycleNodeRequestTransitioner) performInitialHealthChecks(kubeNodes []corev1.Node) error {
+	t.cycleNodeRequest.Status.HealthChecks = make(map[string]v1.HealthCheckStatus)
+
+	// Build a set of ready nodes from which to check below
+	var readyNodesSet = make(map[string]corev1.Node)
```
Jacobious52, line 91:
```suggestion
	readyNodesSet := make(map[string]corev1.Node)
```

pkg/controller/cyclenoderequest/transitioner/healthchecks.go
```diff
@@ -0,0 +1,222 @@
...
+		// All nodes present in the nodegroup before cycling should be skipped after cycling has begun
+		t.cycleNodeRequest.Status.HealthChecks[nodeName] = v1.HealthCheckStatus{
+			Skip: true,
+		}
+	}
+
+	// Perform healthchecks on all nodes in the nodegroup before cycling begins
+	// If any healthchecks fail, cycling should not start
+	for _, node := range t.cycleNodeRequest.Status.NodesToTerminate {
+		nodeName := fmt.Sprintf("%s/%s", node.ProviderID, node.Name)
```
Jacobious52, line 113: this hash function is used multiple times across functions. Break out into its own function so the hash encode format is in one place

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -376,6 +383,26 @@ func (t *CycleNodeRequestTransitioner) transitionScalingUp() (reconcile.Result,
 		}
 	}
 
+	// Skip looping through nodes if no health checks need to be performed
+	if t.cycleNodeRequest.Spec.HealthChecks == nil {
```
Jacobious52, line 387:
```suggestion
	if len(t.cycleNodeRequest.Spec.HealthChecks) == 0 {
```
I think is the more Go approach. Checks both nil and empty

pkg/controller/cyclenodestatus/controller.go
```diff
@@ -114,6 +115,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	rm := cyclecontroller.NewResourceManager(
 		r.mgr.GetClient(),
 		r.rawClient,
+		*http.DefaultClient,
```
Jacobious52, line 118: I was going to say don't use DefaultClient because it has no timeout set, but looks like it's not used for the cns controller here. I see you've set the timeout on her cnr client 👍

However as the client is optional, not used for cns, it might be better to have it as a pointer as pass nil when not expected. Alternatively just make a client with a timeout here too so there's no gotchas in the future.

Jacobious52 (approved):
changes lgtm

dtnyn (approved):

PR #38, opened 2021-11-08
Stability and health check fixes

vincentportella:
- Added a check in the `Initialised` phase to ensure a node is part of the ASG before getting it's nodegroup name
- Return the body of the health check as a string in the cyclops failed cycle error message
- If multiple errors are returned, separate them with a comma


dtnyn (approved):
LGTM

awprice (approved):

PR #40, opened 2021-11-16
workflow: Build arm64 Docker images

### LINKED ISSUE #24: Build arm64 container images in release
Currently there is only one amd64 image inside the release. It would be nice to have arm64 image as well and even better if there is an extra container manifest list for each release to include both two architecture images.

hligit:
Fixes #24


mwhittington21 (approved):

mwhittington21 (approved):

vincentportella (approved):
lgtm

PR #45, opened 2022-02-02
Add pre-termination checks

vincentportella:
- Add node pre-termination checks
- Add `skipInitialHealthChecks` to the nodegroup spec


dtnyn (dismissed):
changes LGTM

pkg/controller/cyclenoderequest/controller.go
```diff
@@ -85,6 +85,30 @@ func NewReconciler(
...
+// returns an error if these are misconfigured
+func tlsCertsValid(tlsConfig v1.TLSConfig) error {
+	// Check that either both the the tls cert and private key are added or missing
+	// If they are both added, cyclops will forward them when making requests for mTLS
+	// If they are both missing, they will not be added. No mTLS.
+	rootCA := os.Getenv(tlsConfig.RootCA)
+	certificate := os.Getenv(tlsConfig.Certificate)
+	key := os.Getenv(tlsConfig.Key)
+
+	if (certificate == "" && key != "") || (certificate != "" && key == "") {
```
Jacobious52, line 98: feels like this can be simplified. essentially it just makes sure that all fields are present?, but with good error messages.

```go
switch {
case certificate == "":
  return fmt.Errorf("certificate is missing")
case key == "":
  return fmt.Errorf("key is missing")
case rootCA == "":
  return fmt.Errorf("rootCA is missing")
default:
  return nil
} 
```

vincentportella: So not quite. The root ca can be present without the cert and key. This would be standard tls. The cert and key can optionally be added together which would the host on the node to authenticate Cyclops via mutual TLS.

cmd/manager/main.go
```diff
@@ -105,8 +105,7 @@ func main() {
 
 	// Setup the notifier if it is enabled
 	if *messagingProviderName != "" {
-		notifier, err = notifierbuilder.BuildNotifier(*messagingProviderName)
-		if err != nil {
+		if notifier, err = notifierbuilder.BuildNotifier(*messagingProviderName); err != nil {
```
awprice, line 108: Hmm not sure about this simplification - even though it works, it might be a little confusing.

If you do `a, err := someFunction(); err != nil {`, `a` and `err` are scoped to that if statement. But if the variables exist and you do `a, err = someFunction(); err != nil {`, (note the `=`) they won't be scoped to the if statement.

pkg/controller/cyclenoderequest/controller.go
```diff
@@ -85,6 +85,30 @@ func NewReconciler(
...
+		return fmt.Errorf("cert or key missing, ensure both neither are missing for mTLS")
```
Jacobious52, line 99: ensure ~both~ neither

cmd/manager/main.go
vincentportella: Sure, I didn't think about how it could cause confusion. In this case I didn't want to scope the the if statement. I'll revert this.

pkg/controller/cyclenoderequest/controller.go
Jacobious52: ahh I see. Essentially there's 3 valid "modes" then?
 
1. No Fields - No TLS
2. CA Only - TLS
3. All Fields - mTLS

vincentportella: Yes exactly. I'll add this as part of the comment to make it clearer.

Jacobious52 (dismissed):
changes lgtm

dtnyn (dismissed):

dtnyn (approved):

Jacobious52 (approved):

PR #47, opened 2022-03-14
Verify minimum compatible client version before cycling a node

bianchi2:
This PR adds a few modifications to client and cyclops controller logic, namely:

* local client will create cnr with an annotation client.api.version=$version
* cyclops controller will read cnr metadata and if exists compare it to a its own client.api.version version
* if there's version mismatch crn status is set to failed and a message is added too
* observer controller creates cnrs with an annotation value retrieved a var that gets its value at build time

Example CNR from a local test where cycle cli was built with a 1.7.0 version:

```
apiVersion: atlassian.com/v1
kind: CycleNodeRequest
metadata:
  annotations:
    client.api.version: 1.0.0
    reason: cli
  generateName: system-
  generation: 2
  name: system-dk5bk
  namespace: kube-system
spec:
  cycleSettings:
    concurrency: 1
    labelsToRemove:
    - coredns-active
    method: Drain
  nodeGroupName: ""
  nodeGroupsList:
  - system.x.x.x.x.x
  selector:
    matchLabels:
      customer: kitt
      role: node
status:
  message: Client API version 0.0.1 does not match controller API version 1.0.0
  phase: Failed
```
Tested cns created by observer. The CNR is created with an annotation which is then validated by cyclops controller.

```
apiVersion: atlassian.com/v1
kind: CycleNodeRequest
metadata:
  annotations:
    client.api.version: 1.0.0
    reason: |-
      instance "i-0be5c478d246b2760" / node "ip-x.x.x.x" not up to date with current cloud provider node group configuration/template
 ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 VERSION = 1.7.1
+# IMPORTANT! Update this version if changes to controller affect objects created by cli
+MINIMAL_COMPATIBLE_CLIENT_VERSION = 1.7.1
 IMAGE = cyclops:$(VERSION)
 ARCH=$(if $(TARGETPLATFORM),$(lastword $(subst /, ,$(TARGETPLATFORM))),amd64)
+BASE_PACKAGE = github.com/atlassian-labs/cyclops/pkg
+OBSERVER_BUILD_LD_FLAGS = -X 'main.version=${VERSION}' -X '${BASE_PACKAGE}/observer.version=${VERSION}'
+MANAGER_BUILD_LD_FLAGS = -X 'main.version=${VERSION}' -X '${BASE_PACKAGE}/controller/cyclenoderequest.minimalCompatibleClientVersion=${MINIMAL_COMPATIBLE_CLIENT_VERSION}'
 
 MANAGER_BIN = cyclops
 CLI_BIN = kubectl-cycle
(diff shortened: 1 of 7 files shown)


Jacobious52 (dismissed):
lgtm. nice feature

Jacobious52 (approved):

vincentportella (approved):

patrickshan:
Change looks good. Just thinking how we are going to do the future release after this to avoid the original problem we had. So basically we need to update both `MINIMAL_COMPATIBLE_CLIENT_VERSION` and `VERSION` when making a PR which has breaking changes before cutting a release to make this work ?

bianchi2:
@patrickshan yes, if changes to cli and controller aren't backward compatible for older cli versions, then MINIMAL_COMPATIBLE_CLIENT_VERSION should be updated.

patrickshan:
@bianchi2 as per our discussion on Slack, `VERSION` will also need to be updated to make sure current master could build a working cli binary.

patrickshan:
lgtm.

patrickshan (approved):

Jacobious52 (approved):

vincentportella (approved):
lgtm

PR #48, opened 2022-04-20
make pre-termination checks concurrent

MinyiZ:
Previously, pre-termination checks in cyclops did not occur concurrently, they ran in sequence when multiple nodes are cycled together.

Now they are run concurrently.

Tested in lab.

Also bump version of golangci-lint so that it uses the correct version of go and not run into [this error](https://github.com/atlassian-labs/cyclops/runs/6104105327?check_suite_focus=true)


vincentportella (approved):
lgtm

dtnyn (approved):

Jacobious52 (approved):

PR #49, opened 2022-04-21
Skip cycling if nodes are up to date with ASG

bianchi2:
This PR:

* adds a flag `--force-asg-config-check` to kubectl cycle which adds `force-check-asg-config` annotation to a cnr
* instructs controller to check for an annotation (true) and check if nodes in the nodegroups are up to date with its ASG (get autoscaling group nodes and call `OutOfDate()` on them)
* if at least one node isn't up to date with ASG cycling will happen as usual
* if all nodes are up to date a cnr is transitioned into a Skipped state with a custom message saying that no node changes detected

Tested in a lab with the following scenarios

### Nodes are up to date with ASG

`kubectl cycle system --force-asg-config-check`. Results in a cnr in a Skipped state:

```
apiVersion: atlassian.com/v1
kind: CycleNodeRequest
metadata:
  annotations:
    client.api.version: 1.0.0
    force-check-asg-config: "true"
    reason: cli
  creationTimestamp: "2022-04-21T21:25:18Z"
  generateName: system-
  generation: 3
  name: system-tdw99
  namespace: kube-system
  resourceVersion: "4286259"
  uid: 06b2a92f-df7a-4d1a-96f5-fe7d558c2024
spec:
  cycleSettings:
    concurrency: 1
    labelsToRemove:
    - coredns-active
    method: Drain
  nodeGroupName: ""
  nodeGroupsList:
  - system.lab-lab5.us-west-2.kitt-inf.net
  selector:
    matchLabels:
      customer: kitt
      role: node
status:
  message: Cycling skipped because all nodes in the nodegroups are up-to-date with
    a cloud provider ASG
  phase: Skipped
```

### Nodes are not up to date with ASG

Change node init script and terraform apply to update system ASG. `kubectl cycle system --force-asg-config-check` results in all system nodes being cycled since there are changes.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitioner.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitioner.go
@@ -91,6 +91,7 @@ func (t *CycleNodeRequestTransitioner) transitionFuncs() map[v1.CycleNodeRequest
 		v1.CycleNodeRequestWaitingTermination: t.transitionWaitingTermination,
 		v1.CycleNodeRequestFailed:             t.transitionFailed,
 		v1.CycleNodeRequestSuccessful:         t.transitionSuccessful,
+		v1.CycleNodeRequestSkipped:			   t.transitionSkipped,
 		v1.CycleNodeRequestHealing:            t.transitionHealing,
 	}
 }
--- a/pkg/controller/cyclenoderequest/transitioner/transitioner.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitioner.go
@@ -91,7 +91,7 @@ func (t *CycleNodeRequestTransitioner) transitionFuncs() map[v1.CycleNodeRequest
 		v1.CycleNodeRequestWaitingTermination: t.transitionWaitingTermination,
 		v1.CycleNodeRequestFailed:             t.transitionFailed,
 		v1.CycleNodeRequestSuccessful:         t.transitionSuccessful,
-		v1.CycleNodeRequestSkipped:			   t.transitionSkipped,
+		v1.CycleNodeRequestSkipped:            t.transitionSkipped,
 		v1.CycleNodeRequestHealing:            t.transitionHealing,
 	}
 }
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -22,6 +22,9 @@ var (
 	nodeEquilibriumWaitLimit = 5 * time.Minute
 )
 
+const ForceAsgConfigCheckAnnotation = "force-check-asg-config"
+
+
 // transitionUndefined transitions any CycleNodeRequests in the undefined phase to the pending phase
 // It checks to ensure that a valid selector has been provided.
 func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result, error) {
@@ -106,6 +109,23 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
 
 	// get instances inside cloud provider node groups
 	nodeGroupInstances := nodeGroups.Instances()
+
+	forceCheckAsgConfigAnnotations := t.cycleNodeRequest.Annotations[ForceAsgConfigCheckAnnotation]
+	if forceCheckAsgConfigAnnotations == "true" {
+		var skipped = true
+		for _, instance := range nodeGroupInstances {
+			outOfDate := instance.OutOfDate()
+			if outOfDate {
+				skipped = false
+				break
+			}
+		}
+		if skipped {
+			t.cycleNodeRequest.Status.Message = "Cycling skipped because all nodes in the nodegroups are up-to-date with a cloud provider ASG"
+			return t.transitionObject(v1.CycleNodeRequestSkipped)
+		}
+	}
+
 	// Do some sanity checking before we start filtering things
 	// Check the instance count of the node group matches the number of nodes found in Kubernetes
 	if len(kubeNodes) != len(nodeGroupInstances) {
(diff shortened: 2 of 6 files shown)


pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -106,6 +109,19 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
...
+
+	forceCheckAsgConfigAnnotations := t.cycleNodeRequest.Annotations[ForceAsgConfigCheckAnnotation]
+	if forceCheckAsgConfigAnnotations == "true" {
+		for _, instance := range nodeGroupInstances {
+			outOfDate := instance.OutOfDate()
+			if outOfDate {
+				break
+			}
+		}
+		t.cycleNodeRequest.Status.Message = "Cycling skipped because all nodes in the nodegroups are up-to-date with a cloud provider ASG"
```
Jacobious52, line 121: does this get skipped when all instance are up to date? the break and completing the loop will both run this code?

bianchi2: Ouch, right. This isn't how it's been tested in a lab. Let me push the original version

bianchi2: @Jacobious52 I have introduce a bool skipped which is set to false once at least one out of date node is identified and then check it to decide if or not to transition a cnr to a Skipped state. Works fine is a lab now.

bianchi2: Fixed by adding a skipped bool which is true by default and set to false if at least one node is not up to date. Tested in a lab by updating asg launch template, deleting a node, letting ASG recreate it so that I have 1 up to date and 2 outdated nodes.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -106,6 +109,23 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
...
 	// get instances inside cloud provider node groups
 	nodeGroupInstances := nodeGroups.Instances()
+
+	forceCheckAsgConfigAnnotations := t.cycleNodeRequest.Annotations[ForceAsgConfigCheckAnnotation]
+	if forceCheckAsgConfigAnnotations == "true" {
+		var skipped = true
+		for _, instance := range nodeGroupInstances {
+			outOfDate := instance.OutOfDate()
+			if outOfDate {
+				skipped = false
```
Jacobious52, line 119: what you could do instead of a bool is just have the return directly in the outOfDate case;

```go
if outOfDate {
  t.cycleNodeRequest.Status.Message = "Cycling skipped because all nodes in the nodegroups are up-to-date with a cloud provider ASG"
  return t.transitionObject(v1.CycleNodeRequestSkipped)
}
```

Will keep the rest of the function after clean

Jacobious52 (dismissed):
lgtm

pkg/controller/cyclenoderequest/transitioner/transitioner.go
```diff
@@ -91,6 +91,7 @@ func (t *CycleNodeRequestTransitioner) transitionFuncs() map[v1.CycleNodeRequest
 		v1.CycleNodeRequestWaitingTermination: t.transitionWaitingTermination,
 		v1.CycleNodeRequestFailed:             t.transitionFailed,
 		v1.CycleNodeRequestSuccessful:         t.transitionSuccessful,
+		v1.CycleNodeRequestSkipped:			   t.transitionSkipped,
```
vincentportella, line 94: nit: align this with the others

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -106,6 +109,23 @@ func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, er
 
 	// get instances inside cloud provider node groups
 	nodeGroupInstances := nodeGroups.Instances()
+
+	forceCheckAsgConfigAnnotations := t.cycleNodeRequest.Annotations[ForceAsgConfigCheckAnnotation]
+	if forceCheckAsgConfigAnnotations == "true" {
```
vincentportella, line 114: Use `strconv.ParseBool()` which will parse a lot of different possible true/false inputs

pkg/controller/cyclenoderequest/transitioner/transitioner.go
bianchi2: Goland + tab weirdness, and indeed git status shows wrong indentation while it's aligned ok in the editor. Should be ok now.

pkg/controller/cyclenoderequest/transitioner/transitions.go
bianchi2: Goog point, even though it's cycle that adds this annotation and its value is controlled, and we don't expect users to add those annotations manually with values like `True`. Changes committed.

vincentportella: imo still handle the error, e.g. if there's an error just assume it's false.

vincentportella: hmm actually that would be the case, ignore my previous statement

vincentportella (approved):

patrickshan:
The original idea is to check if there are any updates (including both ASG config change and Daemonsets config change) before creating a CNR object. The check function should be similar to https://github.com/atlassian-labs/cyclops/blob/master/pkg/observer/controller.go#L117 . At the moment this PR doesn't seem to cover the k8s chagnes, i.e. daemonsets config change part. It would be nice if we could use the same logic as observer when checking the necessity of cycling.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -22,6 +23,9 @@ var (
 	nodeEquilibriumWaitLimit = 5 * time.Minute
 )
 
+const ForceAsgConfigCheckAnnotation = "force-check-asg-config"
```
patrickshan, line 26: nit: add `atlassian.com/cyclops/` prefix for this annotation

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -557,3 +578,34 @@ func (t *CycleNodeRequestTransitioner) transitionSuccessful() (reconcile.Result,
 	t.rm.Logger.Info("Requeuing CycleNodeRequest for deleting later")
 	return reconcile.Result{Requeue: true, RequeueAfter: t.options.DeleteCNRRequeue}, nil
 }
+
+// transitionSkipped handles skipped CycleNodeRequests
+func (t *CycleNodeRequestTransitioner) transitionSkipped() (reconcile.Result, error) {
```
patrickshan, line 583: there seems to be no difference between `transitionSkipped` and `transitionSuccessful`. Any reason not to use  `transitionSuccessful` directly ?

bianchi2:
Closing this one since it does not address the original feature request, and making observer logic means quite a bit of refactoring.

PR #50, opened 2022-08-10
cordon node before performing pre-termination checks

MinyiZ:
- Perform pre-termination checks after cordoning a node instead of before.
- bump linter version and fix minor linter error

Tested in lab 👍


Jacobious52 (approved):
lgtm 💯

hligit (approved):

PR #52, opened 2022-10-07
fix race condition

h8y3:
This PR fixes race conditions where nodes are terminated (i.e by cluster-autoscaler) while in the progress of cycling nodes, which is identify in:

* [Cordon Phase](https://github.com/atlassian-labs/cyclops/blob/156a782add8ce211f1fd8328be09265c99f6b919/pkg/controller/cyclenoderequest/transitioner/transitions.go#L425) - Trying to cordon a node that has been removed from k8s api, CNR failed with

```bash
Node "xxxx" not found
```

* [Healing Phase](https://github.com/atlassian-labs/cyclops/blob/156a782add8ce211f1fd8328be09265c99f6b919/pkg/controller/cyclenoderequest/transitioner/transitions.go#L512) - Trying to re-attach a ec2 instance to ASG that's in terminated/shutting down

To simulate the senario we can insert below to each of the transition phases
```go
// simulate unexpected termination 
for _, node := range t.cycleNodeRequest.Status.CurrentNodes {
err := t.rm.CloudProvider.TerminateInstance(node.ProviderID)
t.rm.LogEvent(t.cycleNodeRequest, "TestTerminate", "Terminating: %v, err: %v", node.Name, err)
_, err = t.rm.GetNode(node.Name)
	}
```

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -376,7 +376,8 @@ func (t *CycleNodeRequestTransitioner) transitionScalingUp() (reconcile.Result, 
 		for i, node := range t.cycleNodeRequest.Status.CurrentNodes {
 			if nodeToRemove.Name == node.Name {
 				t.rm.LogEvent(t.cycleNodeRequest, "RaceCondition", "Node %v was prematurely terminated.", node.Name)
-				t.cycleNodeRequest.Status.CurrentNodes = append(t.cycleNodeRequest.Status.CurrentNodes[:i], t.cycleNodeRequest.Status.CurrentNodes[i+1:]...)
+				t.cycleNodeRequest.Status.CurrentNodes = append(t.cycleNodeRequest.Status.CurrentNodes[:i],
+					t.cycleNodeRequest.Status.CurrentNodes[i+1:]...)
 				break
 			}
 		}
@@ -437,16 +438,18 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result, 
 			// Try to send the trigger, if is has already been sent then this will
 			// be skipped in the function. The trigger must only be sent once
 			if err := t.sendPreTerminationTrigger(node); err != nil {
-				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationTriggerFailed", "failed to send pre-termination trigger to %v, err: %v", node.Name, err)
-				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger to %s is still cordononed", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest,
+					"PreTerminationTriggerFailed", "failed to send pre-termination trigger to %s, err: %v", node.Name, err)
+				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger to %s", node.Name))
 			}
 
 			// After the trigger has been sent, perform health checks to monitor if the node
 			// can be terminated. If all checks pass then it can be terminated.
 			allHealthChecksPassed, err := t.performPreTerminationHealthChecks(node)
 			if err != nil {
-				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationHealChecks", "failed to perform pre-termination health checks to %v, err: %v", node.Name, err)
-				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks to %s", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationHealChecks",
+					"failed to perform pre-termination health checks for %v, err: %v", node.Name, err)
+				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks for %s", node.Name))
 			}
 
 			// If not all health checks have passed, it is not ready for termination yet


vincentportella:
Change the version in the makefile to 1.8.2

dtnyn:
> Change the version in the makefile to 1.8.2

We can wait to cut release in a separate PR to not mix this issue changes with the release cut trigger

vincentportella:
You are correct 🤦‍♂️

pkg/cloudprovider/aws/aws.go
```diff
@@ -50,6 +50,20 @@ func verifyIfErrorOccured(apiErr error, expectedMessage string) (bool, error) {
 	return false, apiErr
 }
 
+func verifyIfErrorOccuredWithDefaults(apiErr error, expectedMessage string) (bool, error) {
+	skip_errs := []string{
+		"is not in correct state",
+	}
+	if awsErr, ok := apiErr.(awserr.Error); ok {
```
vincentportella, line 57: Can you bake this functionality into the `verifyIfErrorOccured` function? Atm you're duplicating functionality. The equivalent can be achieved using a variadic argument, or just by providing a list of expected error messages. In this function you can simply declare them all and pass them to it, this function just happens to provide the specified defaults.

vincentportella: Also it's worth taking the opportunity to fix the typo in the function name, `verifyIfErrorOccured` -> `verifyIfErrorOccurred `

pkg/k8s/node.go
```diff
@@ -40,6 +41,10 @@ func UncordonNode(name string, client kubernetes.Interface) error {
 func IsCordoned(name string, client kubernetes.Interface) (bool, error) {
 	node, err := client.CoreV1().Nodes().Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
+		// considered node is cordoned when it's removed from kube api
+		if apierrors.IsNotFound(err) {
```
vincentportella, line 45: Imo it should be up to the called to decide what to do with the error, not the function itself. When handling the error returned before making pre-termination checks, that is where we can parse the error that has been returned.

vincentportella: I see it more as errors needing to be handled at the top-level rather than in functions. This function is just meant to be a wrapper for making the api call and shouldn't be opinionated about the error returned, that should be up to the caller. A different caller may want to handle the error differently, and adding this functionality to the function itself may block that separate use-case.

dtnyn (dismissed):

pkg/k8s/node.go
vincentportella: By putting it lower in the stack the method itself becomes rigid and catering to the existing use-case. What if in the future the function can be used but because it does include that error skipping it cannot be used. The way I see it, the caller is the one that can meaningfully handle the error, not the function itself. Right now the function is simply ignoring the error, which is what the caller would be doing in this case. You're essentially putting in functionality from the caller in the function.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -439,14 +437,14 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
...
-				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger, %s is still cordononed", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationTriggerFailed", "failed to send pre-termination trigger to %v, err: %v", node.Name, err)
 			}
 
 			// After the trigger has been sent, perform health checks to monitor if the node
 			// can be terminated. If all checks pass then it can be terminated.
 			allHealthChecksPassed, err := t.performPreTerminationHealthChecks(node)
 			if err != nil {
-				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks, %s is still cordononed", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationHealChecks", "failed to perform pre-termination health checks to %v, err: %v", node.Name, err)
```
dtnyn, line 447: I can see that the return value of `t.performPreTerminationHealthChecks(node)`, `allHealthChecksPassed`, is still being used as normal. I don't think that it should continue to used as before as, while it returns what you expect right now, I wouldn't expecting a valid return value from a function after it returning an error.

vincentportella: I agree with what @dtnyn is saying here. If an error is returned, we cannot continue with the normal flow, the error needs to be handled. What is this fixing? The reason why we transition to healing here is that cycling should fail if the pre-termination trigger fails, for whatever reason. Cyclops should continue with cycling under this condition because that means the workload on the node won't be alerted.

dtnyn (approved):
LGTM

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -431,14 +437,16 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
 			// Try to send the trigger, if is has already been sent then this will
 			// be skipped in the function. The trigger must only be sent once
 			if err := t.sendPreTerminationTrigger(node); err != nil {
-				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger, %s is still cordononed", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationTriggerFailed", "failed to send pre-termination trigger to %v, err: %v", node.Name, err)
```
vincentportella, line 440: nit: `node.Name` is a string so you should use `%s` like in the line below.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -431,14 +437,16 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
...
+				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger to %s is still cordononed", node.Name))
```
vincentportella, line 441: nit: can you revert `failed to send pre-termination trigger, %s is still cordononed`, the grammar is no longer correct.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -431,14 +437,16 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
...
 			}
 
 			// After the trigger has been sent, perform health checks to monitor if the node
 			// can be terminated. If all checks pass then it can be terminated.
 			allHealthChecksPassed, err := t.performPreTerminationHealthChecks(node)
 			if err != nil {
-				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks, %s is still cordononed", node.Name))
+				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationHealChecks", "failed to perform pre-termination health checks to %v, err: %v", node.Name, err)
```
vincentportella, line 448:
```suggestion
				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationHealChecks", "failed to perform pre-termination health checks for %v, err: %v", node.Name, err)
```

pkg/controller/cyclenoderequest/transitioner/transitions.go
vincentportella, line 441:
```suggestion
				return t.transitionToHealing(errors.Wrapf(err, "failed to send pre-termination trigger, %s is still cordononed", node.Name))
```

pkg/controller/cyclenoderequest/transitioner/transitions.go
vincentportella, line 440:
```suggestion
				t.rm.LogEvent(t.cycleNodeRequest, "PreTerminationTriggerFailed", "failed to send pre-termination trigger to %s, err: %v", node.Name, err)
```

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -431,14 +437,16 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
...
+				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks to %s", node.Name))
```
vincentportella, line 449:
```suggestion
				return t.transitionToHealing(errors.Wrapf(err, "failed to perform pre-termination health checks for %s", node.Name))
```

dtnyn (approved):

vincentportella (approved):
lgtm

PR #54, opened 2023-06-02
fix the order of nested for loop

vcatlassian:
The order of the 2 nested `for` loop could potentially make `numNodesInProgress++` getting executed for more than once - this won't cause issue when concurrency is 1. Changed the order so that it will loop all ready nodes first. 

Also removed `reapedChildren` from `pkg/controller/cyclenoderequest/transitioner/util.go` which is not used. 

Upgraded github action versions - old versions are deprecated and caused lint job failed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,13 +14,13 @@ jobs:
     steps:
 
     - name: Set up Go
-      uses: actions/setup-go@v2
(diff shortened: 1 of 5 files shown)


vincentportella (approved):
lgtm

dtnyn (approved):

PR #55, opened 2023-06-04
Healing after cyclenodestatus failed

vcatlassian:
When cyclenodestatus failed, cyclenoderequest should transit to Healing instead of Failed directly, so that it will try to re-attach and uncordon nodes. 

One problem with the current logic in `transitionHealing` is that some nodes may have already been terminated when it loops `NodesToTerminate`, and this will make it transit to Failed. Changed it to skip if node doesn't exist anymore, so that it will loop all nodes. 

Note: this will need to rebase from https://github.com/atlassian-labs/cyclops/pull/54 to get the updated github action versions, but I feel it'd still be better to have this as a separate PR. 

p.s. Also updated github action versions for docker.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -520,7 +520,7 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
 
 	for _, node := range t.cycleNodeRequest.Status.NodesToTerminate {
 		// nodes in NodesToTerminate may have been terminated, so check if they still exist
-		if nodeExist, err := k8s.NodeExist(node.Name, t.rm.RawClient); err != nil || !nodeExist {
+		if nodeExists, err := k8s.NodeExists(node.Name, t.rm.RawClient); err != nil || !nodeExists {
 			t.rm.LogEvent(t.cycleNodeRequest,
 				"HealingNodes", "Skip healing node: %s, err: %v",
 				node.Name, err)
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -6,12 +6,13 @@ import (
 	"strings"
 	"time"
 
-	v1 "github.com/atlassian-labs/cyclops/pkg/apis/atlassian/v1"
-	"github.com/atlassian-labs/cyclops/pkg/k8s"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "github.com/atlassian-labs/cyclops/pkg/apis/atlassian/v1"
+	"github.com/atlassian-labs/cyclops/pkg/k8s"
+	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


pkg/k8s/node.go
```diff
@@ -86,3 +86,12 @@ func (c *cachedNodeList) List(selector labels.Selector) ([]*v1.Node, error) {
 
 	return nodes, err
 }
+
+// NodeExist checks if a node exists
+func NodeExist(name string, client kubernetes.Interface) (bool, error) {
```
vincentportella, line 91: nit: `NodeExists`

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -517,28 +518,37 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
...
 		alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName)
 		if alreadyAttached {
 			t.rm.LogEvent(t.cycleNodeRequest,
 				"AttachingNodes", "Skip re-attaching instances to nodes group: %v, err: %v",
 				node.Name, err)
-			continue
 		}
 		if err != nil {
-			return t.transitionToFailed(err)
+			t.rm.LogWarningEvent(t.cycleNodeRequest,
```
vincentportella, line 539: Why change this to a log, don't we want cyclops to transitioned to failed immediately? Otherwise we're in a weird state where we don't know if it should proceed or not.

vcatlassian: As I understand, if it `transitionToFailed` here, it will skip all the rest nodes in the loop, which probably is not expected. For example, if the first node in `NodesToTerminate` has already been successfully cycled and terminated, it will get an error here and return, no more nodes will be looped here to re-attach and uncordon.

vincentportella: Right then you should intercept the error and ensure that that is the case. Logging on every possible error will eventually have unintended consequences. We already do something similar here: https://github.com/atlassian-labs/cyclops/blob/master/pkg/cloudprovider/aws/aws.go#L21

Try to implement some version of that logic in here to ensure it only logs out when you expect it to, otherwise transition to failed because that's not an expected outcome of the request.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -517,28 +518,37 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
 		return t.transitionToFailed(err)
 	}
 
-	// try and re-attach the nodes, if any were un-attached
 	for _, node := range t.cycleNodeRequest.Status.NodesToTerminate {
+		// nodes in NodesToTerminate may have been terminated, so check if they still exist
+		if nodeExists, err := k8s.NodeExists(node.Name, t.rm.RawClient); err != nil || !nodeExists {
```
vincentportella, line 523: You need to catch the error from here. In the function it already checks if the error returned indicated that the node no longer exists which means that legitimate api errors will be returned here. Those need to be caught.

pkg/controller/cyclenoderequest/transitioner/transitions.go
vcatlassian: Cool, I've updated it so that it will return on errors.

dtnyn (approved):
LGTM

PR #59, opened 2023-06-15
Fix if statement in transitionHealing

vcatlassian:
`AttacheInstance` returns `true` and the `"is already part of AutoScalingGroup"` error (not `nil`) when the instance is already in ASG, need to fix the if statement for this - so that when the instance is already in ASG, we ignore this error and continue healing, if there is error other than this, return with error.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -534,12 +534,14 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
 		// try and re-attach the nodes, if any were un-attached
 		t.rm.LogEvent(t.cycleNodeRequest, "AttachingNodes", "Attaching instances to nodes group: %v", node.Name)
 		// if the node is already attached, ignore the error and continue to un-cordoning, otherwise return with error
-		if alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName); alreadyAttached {
+		alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName)
+		if err != nil && !alreadyAttached {
+			return t.transitionToFailed(err)
+		}
+		if alreadyAttached {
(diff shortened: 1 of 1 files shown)


dtnyn (approved):

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -533,13 +533,12 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
 
 		// try and re-attach the nodes, if any were un-attached
 		t.rm.LogEvent(t.cycleNodeRequest, "AttachingNodes", "Attaching instances to nodes group: %v", node.Name)
-		alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName)
-		if alreadyAttached {
+		// if the node is already attached, ignore the error and continue to un-cordoning, otherwise return with error
+		if alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName); alreadyAttached {
```
awprice, line 537: Personally, I find this a little more confusing than what we had before. I think we can still achieve what you're wanting to achieve by not putting the `alreadyAttached, err := nodeGroups.AttachInstance(node.ProviderID, node.NodeGroupName)` into an if statement, but by just adjusting the if statements

vincentportella (approved):

pkg/controller/cyclenoderequest/transitioner/transitions.go
vincentportella: What do you think it updating `nodeGroups.AttachInstance` instead by not returning the error. It was already deemed as being an error that we expected so it would also just not be returned.

vcatlassian: I did think about that, it uses `verifyIfErrorOccurred` to check and return error - which is also used by `DetachInstance`, so I didn't do it. 

Just updated as @awprice suggested, it looks better now I think.

awprice (approved):

vincentportella (approved):

PR #64, opened 2023-09-07
Add a finalizer before detaching a node and don't error if a node has been deleted

vincentportella:
If another controller managed nodes and deletes a node from the cluster during a cycle when this causes Cyclops fail the CNR. There are two improvements made here to mitigate this:
1. Add a finalizer before detaching the node so that the node should not be deleted by another controller
2. Add error checks so that if a node is already deleted from the cluster then cyclops can continue and terminate it

I tested adding the checks when terminating an instance from an ASG but could not produce an error so I left it as is.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -19,10 +19,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-var (
+const (
 	scaleUpWait              = 1 * time.Minute
 	scaleUpLimit             = 20 * time.Minute
 	nodeEquilibriumWaitLimit = 5 * time.Minute
+
+	nodeFinalizerName = "cyclops.atlassian.com/finalizer"
 )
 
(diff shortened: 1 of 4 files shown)


awprice:
We should have an issue describing the problem and have it fixed by this PR

Jacobious52 (approved):
code lgtm.

though, looks like we only remove finalizer on healing state. Are you able to confirm that all transitions to both successful and failed, go through healing to get the finalizer removed? From the state diagram, looks like successful doesn't go to healing.

Would it make more sense to remove the finalizer from within the CNS? right before we do the delete call?

vincentportella:
> code lgtm.
> 
> though, looks like we only remove finalizer on healing state. Are you able to confirm that all transitions to both successful and failed, go through healing to get the finalizer removed? From the state diagram, looks like successful doesn't go to healing.
> 
> Would it make more sense to remove the finalizer from within the CNS? right before we do the delete call?

You're right, I missed that. I'll update the CNS to remove the finalizer from there as well.

Jacobious52 (approved):

vcatlassian (approved):

PR #66, opened 2023-09-14
bump client-go to v0.25.11

dtnyn:
Bumping go-client version to v0.25.11


pkg/cloudprovider/aws/builder.go
```diff
@@ -30,7 +30,7 @@ func NewCloudProvider(logger logr.Logger) (cloudprovider.CloudProvider, error) {
 
 	var log = defaultLogger
 	if logger != nil {
```
awprice, line 32: Possible to remove this logic? Make it so that you must provide a non-pointer logger to `NewCloudProvider`? That way we don't need the default logger logic. Should simplify things

awprice: Will also remove all the places you've had to add `&logger`

dtnyn: Yeah that works too, just wasn't sure if this removal of this logic was desired.
Much simpler not having it around

pkg/apis/atlassian/v1/cyclenoderequest_types.go
```diff
@@ -181,8 +181,9 @@ type CycleNodeRequest struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   CycleNodeRequestSpec   `json:"spec,omitempty"`
-	Status CycleNodeRequestStatus `json:"status,omitempty"`
+	ClusterName string                 `json:"clusterName,omitempty"`
```
awprice, line 184: Was the addition of `ClusterName` intended?

dtnyn: Yeah apimachinery removed the field in v0.25.x https://github.com/kubernetes/apimachinery/blob/v0.25.11/pkg/apis/meta/v1/types.go#L258C34-L260
The field was removed since it wasn't needed by kubernetes
>This field is not set anywhere right now and apiserver is going to ignore it if set in create or update request.

While we don't need it for any kubernetes functionality, we still need it since we piggybacked the meta clusterName field for cyclops notification components.

awprice: Would this place the field inside the `metadata` field though?

dtnyn: No it will be just under `clustername` sine we will (un)marshal it the same way where it's used by the notification https://github.com/atlassian-labs/cyclops/blob/master/pkg/notifications/slack/slack.go#L80-L91. If for convention, we can just create a new struct with the ObjectMeta promoted fields and clustername so we can put it under `metadata`

Jacobious52 (approved):

MinyiZ (approved):

vcatlassian (approved):

vincentportella (approved):

PR #83, opened 2024-07-19
Skip named nodes in the CNR that don't exist

vincentportella:
- Updated to skip named nodes in the `CNR` if they don't exist through a `validationOptions.skipNamedNodes` option. As a result if none of the named nodes exist the `CNR` will succeed immediately when it transitions to the `Initialized` phase.
- `ValidationOptions` option added to the `CNR` and `Nodegroup` objects.
- Very minor refactor of the `Pending` phase and supporting functions.


mwhittington21 (changes requested):
You've significantly altered the behaviour of the software, but haven't added tests for the new behaviour. I would expect to see some tests added that have nodes added to the initial request that are then terminated, and the behaviour validated on that. I'm a little uneasy releasing this functionality without this test.

Additionally, there may be other people running Cyclops who are relying on the existing behaviour. I can see two options here:
1. Cut a new major version of Cyclops (i.e. backwards incompatible changes)
2. Add a configurable setting for this behaviour and default it to the old behaviour

pkg/controller/cyclenoderequest/transitioner/node.go
```diff
@@ -94,33 +104,32 @@ func (t *CycleNodeRequestTransitioner) getNodesToTerminate(numNodes int64) (node
...
-// Returns an error if any named node does not exist in the node group for this CycleNodeRequest.
-func (t *CycleNodeRequestTransitioner) addNamedNodesToTerminate(kubeNodes []corev1.Node, nodeGroupInstances map[string]cloudprovider.Instance) error {
-	for _, namedNode := range t.cycleNodeRequest.Spec.NodeNames {
-		foundNode := false
-		for _, kubeNode := range kubeNodes {
-			if kubeNode.Name == namedNode {
-				foundNode = true
+// Skips any named node that does not exist in the node group for this CycleNodeRequest.
+func (t *CycleNodeRequestTransitioner) addNamedNodesToTerminate(kubeNodes map[string]corev1.Node, nodeGroupInstances map[string]cloudprovider.Instance) {
+	kubeNodesMap := make(map[string]corev1.Node)
```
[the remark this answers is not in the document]
vincentportella: > Additionally, there may be other people running Cyclops who are relying on the existing behaviour.
Good point, I initially intended for this to be a breaking change but now that I think about it making it configurable may be the better approach to cater to different use-cases.

👍 I'll add some tests to validate the intended behaviour.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -64,13 +64,11 @@ func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result,
 func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, error) {
 	// Fetch the node names for the cycleNodeRequest, using the label selector provided
 	t.rm.LogEvent(t.cycleNodeRequest, "SelectingNodes", "Selecting nodes with label selector")
+
 	kubeNodes, err := t.listReadyNodes(true)
 	if err != nil {
 		return t.transitionToHealing(err)
 	}
-	if len(kubeNodes) == 0 {
```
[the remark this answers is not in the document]
vincentportella: I don't really see how that's possible unless we have a way of knowing if a node name in the CNR matches a node that existed prior to the CNR being created.

A flag here would help to toggle strict node validation and enable both use cases but I'm wary of adding more settings to a CNR.

pkg/controller/cyclenoderequest/transitioner/node.go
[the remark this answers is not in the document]
vincentportella: Additionally, there may be other people running Cyclops who are relying on the existing behaviour.

Good point, I initially intended for this to be a breaking change but now that I think about it making it configurable may be the better approach to cater to different use-cases.

👍 I'll add some tests to validate the intended behaviour.

MinyiZ (approved):

deploy/crds/atlassian.com_cyclenodestatuses_crd.yaml
```diff
@@ -101,6 +101,13 @@ spec:
                     - Drain
                     - Wait
                     type: string
+                  strictValidation:
```
awprice, line 104: So by default if not set this will be `false`?

> Added a new StrictValidation option to the CNR to keep backwards compatible functionality.

I guess this does keep backwards compatible functionality, but is technically a breaking change or a behaviour change - we'll need to document this in the release notes

pkg/apis/atlassian/v1/common_types.go
```diff
@@ -46,6 +46,11 @@ type CycleSettings struct {
 	// in-progress CNS request timeout from the time it's worked on by the controller.
 	// If no cyclingTimeout is provided, CNS will use the default controller CNS cyclingTimeout.
 	CyclingTimeout *metav1.Duration `json:"cyclingTimeout,omitempty"`
+
+	// StrictValidation is a boolean which determines whether named nodes selected in a CNR must
+	// exist and be valid nodes before cycling can begin. If set to true when invalid nodes are
+	// selected the CNR will be transitioned to the "Failed" phase before cycling can begin again.
+	StrictValidation bool `json:"strictValidation,omitempty"`
```
awprice, line 53: Do we ever see `StrictValidation` being used for other things? 

At the moment the description of this field suggests its only for this one thing - when invalid nodes are found, transition to failed. But the naming of the field suggests its to generically enable/disable strict validation.

Naming is hard, but I'd probably suggest renaming the flag to something that describes specifically what it is enabling/disabling

awprice (changes requested):

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
```diff
@@ -44,6 +44,8 @@ spec:
               of an object. Servers should convert recognized schemas to the latest
               internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
+          clusterName:
```
awprice, line 47: Where does this field come from? Why is it only in the `CycleNodeRequest` resource?

pkg/apis/atlassian/v1/common_types.go
awprice: Does it also make sense to put this field into `CycleSettings`? 

Doing this means it's available in a CycleNodeStatus object - do we need it here? 

Should we move it elsewhere within CycleNodeRequest to make it not propagate down?

Could we have a "ValidationOptions" field within CycleNodeRequest for holding other validation/configuration options a user might want to configure?

deploy/crds/atlassian.com_cyclenodestatuses_crd.yaml
vincentportella: Yeah that is the intention, I would like this new functionality to be the default rather than the exception because it reduces the number of times a cycle will fail by default.

deploy/crds/atlassian.com_cyclenoderequests_crd.yaml
vincentportella: It was part of the operator-sdk render, `make generate-crds`

awprice (approved):
LGTM - nice work! Love the `validationOptions` field

mwhittington21 (approved):
LGTM

PR #85, opened 2024-08-09
Add mocking to Cyclops

vincentportella:
- Adding a new mocking package to help test individual phases of Cyclops in a repeatable manner. This PR only includes tests for the `Pending` phase and includes the required functionality for that.
- Fixing an issue where if no nodegroups names are given then all nodegroups are determined to be matching.


pkg/cloudprovider/aws/builder.go
```diff
@@ -40,3 +42,11 @@ func NewCloudProvider(logger logr.Logger) (cloudprovider.CloudProvider, error) {
 
 	return p, nil
 }
+
+// NewMockCloudProvider returns a new mock AWS cloud provider
+func NewMockCloudProvider(autoscalingiface autoscalingiface.AutoScalingAPI, ec2iface ec2iface.EC2API) cloudprovider.CloudProvider {
```
awprice, line 47: Is this really returning a new mock AWS cloud provider or just a new provider that could either be the real thing or mocked depending on what is passed in?

pkg/mock/phase_pending_test.go
awprice, line 1: Is this an appropriate location in the codebase for testing the Pending phase? I'd have thought these tests would be better located with the phase implementations themselves.

pkg/mock/phase_pending_test.go
```diff
@@ -0,0 +1,297 @@
...
+					Concurrency: 1,
+					Method:      v1.CycleNodeRequestMethodDrain,
+				},
+				Selector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"customer": "kitt",
+					},
+				},
+				NodeNames: []string{
+					"node-0",
```
awprice, line 85: I guess one nit here is in this test we're assuming that when you create a test node group with `newNodegroup()` that nodes follow this naming scheme. 

Someone modifying `newNodegroup()` may find they break a ton of tests if they decide to change an assumption on how test node groups are created

awprice: You could possibly remedy this by adding a node that obviously would very obviously never exist in the `NodeNames` field

pkg/mock/phase_pending_test.go
```diff
@@ -0,0 +1,297 @@
...
+		WithKubeNodes(nodegroup),
+		WithCloudProviderInstances(nodegroup),
+	)
+
+	// Should requeue while it tries to wait
+	_, err = mockTransitioner.Run()
+	assert.NoError(t, err)
+	assert.Equal(t, v1.CycleNodeRequestPending, mockTransitioner.Cnr.Status.Phase)
+
+	// Simulate waiting for 1 day, this is more than the waiting limit
```
awprice, line 215: Are we assuming the waiting limit is always < 24 hours? Can we make this an input into the test, so that someone changing the wait limit doesn't break unrelated tests

pkg/test/aws.go
```diff
@@ -1,53 +0,0 @@
...
-	TerminateInstanceInAutoScalingGroupErr    error
-}
-
-// DescribeAutoScalingGroups mock implementation for MockAutoscalingService
-func (m MockAutoscalingService) DescribeAutoScalingGroups(*autoscaling.DescribeAutoScalingGroupsInput) (*autoscaling.DescribeAutoScalingGroupsOutput, error) {
-	return m.DescribeAutoScalingGroupsOutput, m.DescribeAutoScalingGroupsErr
-}
-
-// SetDesiredCapacity mock implementation for MockAutoscalingService
-func (m MockAutoscalingService) SetDesiredCapacity(*autoscaling.SetDesiredCapacityInput) (*autoscaling.SetDesiredCapacityOutput, error) {
```
awprice, line 32: I notice we've not implemented functions like this in the new mock AWS - do we not need them?

pkg/cloudprovider/aws/builder.go
vincentportella: Good point, it's not necessarily mocking.

pkg/test/aws.go
vincentportella: I will implement functionality as needed. As I add new features in upcoming PRs I will add the functionality for the mocking to support the appropriate tests.

pkg/mock/phase_pending_test.go
vincentportella: Unfortunately that leads to an import cycle so this is the best place I could think of putting it

mwhittington21: The mock for an individual piece of functionality can live next to the code too, I've seen this happen in quite a few code bases. That avoids the import cycle. You just import generic mocking functionality from the central package in that situation.

pkg/mock/util.go
```diff
@@ -0,0 +1,49 @@
...
+	nodes := make([]*Node, 0)
+
+	for i := 0; i < num; i++ {
+		instanceID, err := generateRandomInstanceId()
+		if err != nil {
+			return nil, err
+		}
+
+		node := &Node{
+			Name:       fmt.Sprintf("node-%d", i),
```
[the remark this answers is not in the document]
vincentportella: Good point, I can add the nodegroup name to the node name to help get around this.

pkg/mock/util.go
```diff
@@ -0,0 +1,49 @@
...
+	hexString := hex.EncodeToString(randomBytes)
+	hexString = hexString[:17]
+	return "i-" + hexString, nil
+}
+
+func newNodegroup(name string, num int) ([]*Node, error) {
+	nodes := make([]*Node, 0)
+
+	for i := 0; i < num; i++ {
+		instanceID, err := generateRandomInstanceId()
```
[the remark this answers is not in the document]
vincentportella: This is more of a helper, instance IDs can be made customer when someone would want to create a node without it.

pkg/mock/phase_pending_test.go
awprice: > Unfortunately that leads to an import cycle so this is the best place I could think of putting it

If this is occurring and as @mwhittington21 suggests we'll need to make the mocking package more generic and moving mocking logic to where it makes sense (probably next to the thing it is mocking)

pkg/cloudprovider/aws/fake/fake.go
```diff
@@ -0,0 +1,142 @@
...
+type Instance struct {
+	InstanceID           string
+	AutoscalingGroupName string
+	State                string
+}
+
+type Ec2 struct {
+	ec2iface.EC2API
+
+	Instances *[]*Instance
```
awprice, line 27: Related to my other comment on pointer to a slice - why?

vincentportella: It was to take into terminating instances using the ec2 instance and then describing asgs using the autoscaling instance and I want them to be in sync. I've done some testing and moved to make to make it easier to manage those fake instances in the future and it works as I'd like.

pkg/cloudprovider/aws/builder.go
```diff
@@ -40,3 +41,19 @@ func NewCloudProvider(logger logr.Logger) (cloudprovider.CloudProvider, error) {
 
 	return p, nil
 }
+
+// NewGenericCloudProvider returns a new mock AWS cloud provider
+func NewGenericCloudProvider(autoscalingiface *fakeaws.Autoscaling, ec2iface *fakeaws.Ec2) cloudprovider.CloudProvider {
+	// autoscalingiface := &fakeaws.Autoscaling{
```
awprice, line 47: Do we need to keep this if it's commented out?

vincentportella: Good catch that was a mistake

awprice (approved):

MinyiZ (approved):

PR #86, opened 2024-08-19
Light Pending phase refactor help with future work to deal with problem nodes

vincentportella:
Small PR to refactor the Pending phase to set the stage for dealing with the different kinds of problem nodes that can occur when collecting the nodes in the Pending phase. This PR does makes minimal changes to the behavior or cyclops. All tests pass as before.

Moving the check for the equilibrium timeout to the start of the Pending phase. This feels like more appropriate placement as it should run before any attempts that are made to fix the state of the nodes in kube and the cloud provider. If it reaches the time to time out then it should fail, not have another attempt at fixing things since the timeout period is long enough at 5 minutes.

### DIFF
--- a/pkg/cloudprovider/aws/fake/fake.go
+++ b/pkg/cloudprovider/aws/fake/fake.go
@@ -111,6 +111,16 @@ func (m *Autoscaling) DescribeAutoScalingGroups(input *autoscaling.DescribeAutoS
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


mwhittington21:
Looks good. The one thing I'd consider is can you think of a test to add that would capture the new behaviour, so we can verify it stays that way over time?

vincentportella:
@mwhittington21 I've added a new test. It succeeds when run with the changes of this PR and I've confirmed that on master the test fails as expected.

mwhittington21 (approved):

awprice (approved):

PR #87, opened 2024-08-22
Fix node state in the Pending phase before cycling

vincentportella:
Updates to the Pending phase:
- Finding all nodes rather than just `Ready` nodes to get an accurate state of the nodes/instances prior to cycling
- Now deleting any node objects in kube that do not have a matching instance in AWS
- Now re-attaching any instances that were detached from the ASG, likely due to an issue in a previous cycle.
- Now when checking if a node has a matching instance in AWS, no longer counting instances that are terminated which would still show up in the list.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/transitions.go
+++ b/pkg/controller/cyclenoderequest/transitioner/transitions.go
@@ -61,19 +61,11 @@ func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result, 
 //  2. describes the node group and checks that the number of instances in the node group matches the number we
 //     are planning on terminating
 func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, error) {
-	// Start the equilibrium wait timer, if this times out as the set of nodes in kube and
-	// the cloud provider is not considered valid, then transition to the Healing phase as
-	// cycling should not proceed.
-	timedOut, err := t.equilibriumWaitTimedOut()
-	if err != nil {
+	// Start the equilibrium wait timer, if this times out then the set of nodes in kube and
+	// the cloud provider is not considered valid. Transition to the Healing phase as cycling
+	// should not proceed.
+	if err := t.errorIfEquilibriumTimeoutReached(); err != nil {
 		return t.transitionToHealing(err)
-	}
-
-	if timedOut {
-		return t.transitionToHealing(fmt.Errorf(
-			"node count mismatch, number of kubernetes nodes does not match number of cloud provider instances after %v",
-			nodeEquilibriumWaitLimit,
-		))
 	}
 
 	// Fetch the node names for the cycleNodeRequest, using the label selector provided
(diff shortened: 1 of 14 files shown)


pkg/controller/node.go
```diff
@@ -96,6 +98,36 @@ func (rm *ResourceManager) DrainPods(nodeName string, unhealthyAfter time.Durati
 	return false, k8s.DrainPods(pods, rm.RawClient, unhealthyAfter)
 }
 
+func (rm *ResourceManager) AddNodegroupAnnotationToNode(nodeName, nodegroupName string) error {
```
awprice, line 101: Is there a reason for using an annotation for the node group rather than a label?

awprice:
Are we also at some point going to update docs with all this new functionality/tweaks to existing functionality you've added recently?

pkg/controller/node.go
vincentportella: Mainly because it's not identifying the CNR itself but rather is used by the controller to link back to the ASG.

awprice: I guess more question was more around - we already have logic for adding a label to a node in the Cyclops codebase - can we re-use it?

vincentportella: Yeah there's nothing stopping us from doing that, if that's preferable happy to go that way. I did it because it was minimal work and more strictly adheres to the definitions from what I understand.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -61,21 +61,13 @@ func (t *CycleNodeRequestTransitioner) transitionUndefined() (reconcile.Result,
 //  2. describes the node group and checks that the number of instances in the node group matches the number we
 //     are planning on terminating
 func (t *CycleNodeRequestTransitioner) transitionPending() (reconcile.Result, error) {
```
mwhittington21, line 63: General note: I'm worried we're making the pending transition a bit too thick with logic. It would be nice if, longer term, we can start to add more states to deal with fixing up a CNR. We could transition from pending -> the state where we are fixing node related problems -> back to pending.

We could also just work on collapsing the total vertical space used by the function and see how we feel afterwards. To me, the function is a bit longer than acceptable at this point. I wouldn't fix this up in this particular PR though.

pkg/controller/cyclenoderequest/transitioner/transitions_pending_test.go
```diff
@@ -293,10 +294,65 @@ func TestPendingNoKubeNodes(t *testing.T) {
...
 }
 
-// Test to ensure that Cyclops will not proceed if there is node detached from
-// the nodegroup on the cloud provider. It should try to wait for the issue to
-// resolve to transition to the Healing phase if it doesn't.
-func TestPendingDetachedCloudProviderNode(t *testing.T) {
+// Test to ensure that node objects that exist in the cluster without a matching
+// instance in the cloud provider are cleaned up and then cycling can proceed
+// as normal.
+func TestPendingOrphanedKubeNodes(t *testing.T) {
```
mwhittington21, line 300: Is there a reason the tests can't all go in one `_test.go` file and just be named and commented appropriately?

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -295,31 +330,202 @@ func findProblemNodes(kubeNodes map[string]corev1.Node, nodeGroupInstances map[s
...
+		// object and ensure that only the Cyclops finalizer exists on it. If another finalizer
+		// exists then it's from another controller and it should not be deleted.
+		containsNonCyclingFinalizer, err := t.rm.NodeContainsNonCyclingFinalizer(node.Name)
+		if err != nil {
+			return errors.Wrap(err,
+				fmt.Sprintf("failed to check if node %s contains a non-cycling finalizer", node.Name),
+			)
+		}
+
+		if containsNonCyclingFinalizer {
```
mwhittington21, line 432: I think if we detect a non cycling finalizer we should probably retry a few times to see if the other finalizer is acting on the object. This should only happen once the Node is actually being deleted though. We can start with this implementation, it seems fine.

Technically, are we misusing finalizers here? We could go all in on proper usage and have a webhook and everything to clean these up properly, but that also starts to impact cluster stability. In that case, should we juse use an annotation and no finalizer?

mwhittington21 (dismissed):

pkg/controller/cyclenoderequest/transitioner/transitions_pending_test.go
vincentportella: I can see a lot of tests being written to different phases, i'd like to not have one mega file with all the tests that becomes hard to navigate.

pkg/controller/cyclenoderequest/transitioner/transitions.go
vincentportella: I agree with you, we should do it but not in this PR. I did toy with doing it that way but I found it would be a bigger change than I wanted.

pkg/controller/cyclenoderequest/transitioner/util.go
vincentportella: > I think if we detect a non cycling finalizer we should probably retry a few times to see if the other finalizer is acting on the object. This should only happen once the Node is actually being deleted though. We can start with this implementation, it seems fine.

If we see issues I'm happy to go in a change it at a later point.

> Technically, are we misusing finalizers here? We could go all in on proper usage and have a webhook and everything to clean these up properly, but that also starts to impact cluster stability. In that case, should we juse use an annotation and no finalizer?

We're using a finalizer to prevent anything from prematurely deleting the node after it has been detached from the ASG. The finalizer is just there to allow cyclops to control the deletion process by drain pods as intended, etc. This is the intended use no?

awprice (approved):

mwhittington21 (approved):

pkg/controller/cyclenoderequest/transitioner/transitions.go
mwhittington21: I agree that change should happen in its own refactoring PR.

pkg/controller/cyclenoderequest/transitioner/util.go
mwhittington21: This is a valid use for a finalizer, agreed. Thanks for clarifying.

PR #88, opened 2024-08-23
Add a max failed CNRs threshold to Nodegroups

vincentportella:
- Adding a new `maxFailedCycleNodeRequests` attribute to the `NodeGroup` spec which defines how many failed CNRs are allowed for a nodegroup before observer will stop generating new ones. The equivalent to the current behaviour is `maxFailedCycleNodeRequests: 0`. This is not a breaking change.
- The Cyclops controller will now delete any sibling failed CNRs when a new one reaches the `Successful` phase.
- The `cyclops_cycle_node_requests_by_phase` metric has been update to include a new `nodegroup` attribute which is the name of the nodegroup the CNR was generated from. Manually created CNRs that don't match a nodegroup will include the attribute with an empty string.

### DIFF
--- a/pkg/controller/cyclenoderequest/transitioner/test_helpers.go
+++ b/pkg/controller/cyclenoderequest/transitioner/test_helpers.go
@@ -26,7 +26,7 @@ func WithKubeNodes(nodes []*mock.Node) Option {
 
 func WithExtraKubeObject(extraKubeObject client.Object) Option {
 	return func(t *Transitioner) {
-		t.extrakubeObjects = append(t.extrakubeObjects, extraKubeObject)
+		t.extraKubeObjects = append(t.extraKubeObjects, extraKubeObject)
 	}
 }
 
@@ -45,7 +45,7 @@ type Transitioner struct {
 	CloudProviderInstances []*mock.Node
 	KubeNodes              []*mock.Node
 
-	extrakubeObjects []client.Object
+	extraKubeObjects []client.Object
 
 	transitionerOptions Options
 }
@@ -56,7 +56,7 @@ func NewFakeTransitioner(cnr *v1.CycleNodeRequest, opts ...Option) *Transitioner
 		// override these as needed
 		CloudProviderInstances: make([]*mock.Node, 0),
 		KubeNodes:              make([]*mock.Node, 0),
-		extrakubeObjects:       []client.Object{cnr},
+		extraKubeObjects:       []client.Object{cnr},
 		transitionerOptions:    Options{},
 	}
 
@@ -65,7 +65,7 @@ func NewFakeTransitioner(cnr *v1.CycleNodeRequest, opts ...Option) *Transitioner
 	}
 
 	t.Client = mock.NewClient(
-		t.KubeNodes, t.CloudProviderInstances, t.extrakubeObjects...,
+		t.KubeNodes, t.CloudProviderInstances, t.extraKubeObjects...,
 	)
 
 	rm := &controller.ResourceManager{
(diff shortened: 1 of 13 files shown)


awprice (approved):

pkg/controller/cyclenoderequest/transitioner/test_helpers.go
```diff
@@ -30,6 +44,10 @@ type Transitioner struct {
 
 	CloudProviderInstances []*mock.Node
 	KubeNodes              []*mock.Node
+
+	extrakubeObjects []client.Object
```
mwhittington21, line 48: nit: extraKubeObjects

pkg/controller/cyclenoderequest/transitioner/transitions_successful_test.go
```diff
@@ -0,0 +1,275 @@
...
+		Spec: v1.CycleNodeRequestSpec{
+			NodeGroupName: "ng-2",
+		},
+		Status: v1.CycleNodeRequestStatus{
+			Phase:   v1.CycleNodeRequestFailed,
+			Message: "",
+		},
+	}
+
+	// Failed CNR for the same nodegroup in a different namespace
```
mwhittington21, line 206: Are CNRs intended to be namespaced? I can't see much value in separating them by namespace like this but being able to act on the same set of backing nodes provided by a cloud provider.

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -536,3 +536,38 @@ func (t *CycleNodeRequestTransitioner) validateInstanceState(validNodeGroupInsta
 
 	return false, nil
 }
+
+// deleteFailedSiblingCNRs finds the CNRs generated for the same nodegroup as
+// the one in the calling transitioner. It filters for deleted CNRs in the same
```
mwhittington21, line 541: nit: it's not a calling transitioner, it's the parent struct technically. I would just say "the one in the transitioner"

pkg/observer/controller.go
```diff
@@ -230,21 +230,42 @@ func (c *controller) inProgressCNRs() v1.CycleNodeRequestList {
...
+			// If the number of Failed CNRs exceeds the threshold in the
+			// nodegroup then drop it
+			if failedCNRsFound > nodeGroup.Spec.MaxFailedCycleNodeRequests {
+				dropNodeGroup = true
 			}
 		}
-		if found {
+
+		if dropNodeGroup {
 			klog.Warningf("nodegroup %q has an in progress CNR.. skipping this nodegroup", nodeGroup.Name)
```
mwhittington21, line 264: This log message now doesn't match reality. dropNodeGroup is being set to true when the number of failed CNRs exceeds the threshold, and not when it has an in progress CNR.

mwhittington21 (changes requested):

pkg/controller/cyclenoderequest/transitioner/transitions_successful_test.go
vincentportella: That would go back to when they were designed. I'm not particularly sure, though changing that is out of scope here.

pkg/observer/controller.go
vincentportella: I will add a second log line there because `Failed` is defined as "in progress" in the existing implementation. I'm adding functionality to skip a certain number of them from counting as "in progress" so I will update for that.

mwhittington21 (approved):

mwhittington21 (dismissed):

mwhittington21 (approved):

MinyiZ (approved):

PR #89, opened 2024-08-23
Change Wait method to select undisruptable pods using annotation

vincentportella:
Changing the `Wait` method to select pods that cannot be drained on a node using an annotation rather than just `DaemonSet` and `Static` pods.

### DIFF
--- a/docs/cycling/README.md
+++ b/docs/cycling/README.md
@@ -106,6 +106,11 @@ spec:
     - "node-name-A"
     - "node-name-B"
 
+  # Optional section - collection of validation options to define stricter or more lenient validation during cycling.
+  validationOptions:
+    # Optional field - Skip node names defined in the CNR that do not match any existing nodes in the Kubernetes API.
+    skipMissingNamedNodes: true|false
+
   cycleNodeSettings:
       # Method can be "Wait" or "Drain", defaults to "Drain" if not provided
       # "Wait" will wait for pods with the "cyclops.atlassian.com/do-not-disrupt=true"
--- a/docs/cycling/README.md
+++ b/docs/cycling/README.md
@@ -35,11 +35,11 @@ The CycleNodeRequest CRD handles a request to cycle nodes belonging to a specifi
 5. In the **ScalingUp** phase, wait for the cloud provider to bring up the new nodes and then wait for the new nodes to be **Ready** in the Kubernetes API. Wait for the configured health checks on the node succeed. Transition the object to **CordoningNode**.
 
 6. In the **CordoningNode** phase, cordon the selected nodes in the Kubernetes API then perform the pre-termination checks. Transition the object to **WaitingTermination**.
-    
+
 7. In the **WaitingTermination** phase, create a CycleNodeStatus CRD for every node that was cordoned. Each of these CycleNodeStatuses handles the termination of an individual node. The controller will wait for a number of them to enter the **Successful** or **Failed** phase before moving on.
-    
+
     If any of them have **Failed** then the CycleNodeRequest will move to **Failed** and will not add any more nodes for cycling. If they are all **Successful** then the CycleNodeRequest will move back to **Initialised** to cycle more nodes.
-    
+
 ### CycleNodeStatus
 
 The CycleNodeStatus CRD handles the draining of pods from, and termination of, an individual node. These should only be created by the controller.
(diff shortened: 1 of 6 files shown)


docs/cycling/README.md
```diff
@@ -106,14 +106,10 @@ spec:
     - "node-name-A"
     - "node-name-B"
 
-  # Optional section - collection of validation options to define stricter or more lenient validation during cycling.
```
awprice, line 109: Is removing these docs intentional?

pkg/k8s/pod.go
```diff
@@ -84,6 +85,25 @@ func PodIsDaemonSet(pod *v1.Pod) bool {
...
+
+// PodCannotBeDisrupted returns true if the pod cannot be forcibly drained from
+// a node.
+func PodCannotBeDisrupted(pod *v1.Pod) bool {
+	for annotationName, annotationValue := range pod.ObjectMeta.Annotations {
+		if annotationName != doNotDisruptAnnotation {
+			continue
+		}
+
+		if annotationValue != "true" {
```
awprice, line 100: Do we care if the value is `true`? Could make this a little bit foolproof by just checking if the annotation key is present.

docs/cycling/examples/README.md
```diff
@@ -166,18 +166,34 @@ spec:
...
+This example shows the usage of the `Wait` method which as opposed to `Drain`, which attempts to remove pods from the node before terminating, will wait for pods with the `cyclops.atlassian.com/do-not-disrupt=true` annotation to leave the node naturally by themselves. This is useful for situations where you cannot forcefully remove pods, such as high churn jobs which need to be run to completion.
+
+```yaml
+# Pod example
+apiVersion: v1
+kind: Pod
+metadata:
+  name: do-not-disrupt
+  annotations:
+    cyclops.atlassian.com/do-not-disrupt: "true"
```
awprice, line 178: Is an annotation appropriate for this? Label would allow both cyclops and cluster users to select un-disruptable pods with a labelSelector, potentially useful in various scenarios.

pkg/k8s/pod.go
awprice: If we do decide to keep it, I'd like to see the expected value to be a const

pkg/k8s/pod.go
```diff
@@ -84,6 +85,25 @@ func PodIsDaemonSet(pod *v1.Pod) bool {
...
 	}
+
+	return false
+}
+
+// PodCannotBeDisrupted returns true if the pod cannot be forcibly drained from
+// a node.
+func PodCannotBeDisrupted(pod *v1.Pod) bool {
+	for annotationName, annotationValue := range pod.ObjectMeta.Annotations {
+		if annotationName != doNotDisruptAnnotation {
```
awprice, line 96: Could possible make this logic even more concise/clear - 
```
if annotationName == doNotDisruptAnnotation && annotationValue == "true" {
    return true
}
```

docs/cycling/README.md
vincentportella: Nope not intentional, putting it back

docs/cycling/examples/README.md
vincentportella: I'm modeling this around how karpenter does it: https://karpenter.sh/docs/concepts/disruption/#pod-level-controls

pkg/k8s/pod.go
vincentportella: I would like to keep it otherwise a situation where `cyclops.atlassian.com/do-not-disrupt: "false"` is possible and to me that's not particularly clear. I'd rather keep it strict and define that in the docs, it doesn't make it any harder to use it.

pkg/k8s/pod.go
```diff
@@ -15,7 +15,9 @@ import (
 )
 
 const (
-	podConditionTypeForUnhealthy = v1.PodReady
+	podConditionTypeForUnhealthy       = v1.PodReady
+	doNotDisruptAnnotation             = "cyclops.atlassian.com/do-not-disrupt"
+	doNotDisruptAnnotationRequredValue = "true"
```
awprice, line 20: Typo in `doNotDisruptAnnotationRequredValue`, should be `doNotDisruptAnnotationRequiredValue`

awprice (approved):
LGTM!

mwhittington21 (approved):

PR #91, opened 2025-01-20
remove safeToStartCycle check

MinyiZ:
This section of code does not behave as intended. It was introduced in 2020 to attempt to fix a race condition between the observer and cluster autoscaler. Due to recent improvements, the race condition is now solved so this code can be removed.

**Why does this code not work?**
From https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/metrics.md#cluster-autoscaler-execution

> `last_activity` records last time certain part of cluster autoscaler logic executed. Represented with unix timestamp. autoscaler-activity values are:
> 
> main - main loop iteration started.
> 
> autoscaling - current state of the cluster has been updated, started autoscaling logic.
> 
> scaleUp - autoscaler will check if scale up is necessary.
> 
> scaleDown - autoscaler will try to scale down some nodes.

The `last_activity` metric reflects whether cluster autoscaler has attempted to scale up or down. However, it does not necessarily indicate actual changes in the ASG, resulting in numerous false positives. This behaviour can unnecessarily block the creation of CNRs, even when it is safe to proceed with cycling.

### DIFF
--- a/cmd/observer/main.go
+++ b/cmd/observer/main.go
@@ -37,7 +37,6 @@ type app struct {
 	cloudProviderName        *string
 	namespace                *string
 	addr                     *string
-	prometheusAddress        *string
 	dryMode                  *bool
 	runImmediately           *bool
 	runOnce                  *bool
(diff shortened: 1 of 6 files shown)


cmd/observer/main.go
```diff
@@ -50,18 +49,16 @@ type app struct {
...
-		waitInterval:             rootCmd.PersistentFlags().Duration("wait-interval", 2*time.Minute, "duration to wait after detecting changes before creating CNR objects. The window for letting changes on nodegroups settle before starting rotation"),
-		checkInterval:            rootCmd.PersistentFlags().Duration("check-interval", 5*time.Minute, `duration interval to check for changes. e.g. run the loop every 5 minutes"`),
-		nodeStartupTime:          rootCmd.PersistentFlags().Duration("node-startup-time", 2*time.Minute, "duration to wait after a cluster-autoscaler scaleUp event is detected"),
-		runImmediately:           rootCmd.PersistentFlags().Bool("now", false, "makes the check loop run straight away on program start rather than wait for the check interval to elapse"),
-		runOnce:                  rootCmd.PersistentFlags().Bool("once", false, "run the check loop once then exit. also works with --now"),
-		prometheusAddress:        rootCmd.PersistentFlags().String("prometheus-address", "prometheus", "Prometheus service address used to query cluster-autoscaler metrics"),
```
mwhittington21, line 63: discuss: is it worth allowing `--prometheus-address` to remain as a flags but be unused, in order to avoid erroring out people's installations when they automatically bump the version, or is it better for the version bump to immediately show this feature isn't used any more by this error message stopping it starting?
```
docker run --rm -it ghcr.io/atlassian-labs/cyclops:v1.10.1 cyclops --prometheus-address 127.0.0.1:8080
cyclops: error: unknown long flag '--prometheus-address', try --help
```

mwhittington21 (approved):

cmd/observer/main.go
MinyiZ: i think i prefer to remove the flag immediately because having a flag that does nothing can be confusing. leaving it would also require us to remember to remove it later. since this is already a breaking change, i think it's better to alert users to the change in behaviour right away with the flag removal, rather than silently ignoring it.

vincentportella (approved):
lgtm

PR #117, opened 2025-08-08
Add priority-based node cycling with dedup, retries, and metrics

dnorman3:
### Summary
Implements staged priority-based node cycling in Cyclops. All CNRs are created upfront for visibility and then activated sequentially by ascending priority once lower priorities complete successfully and pass health checks. 
Adds performance optimisations, retry logic, deduplication, and metrics. 

### Motivation
Prevent cluster degradation by ensuring default node groups are healthy before cycling system node groups.
Provides visibility into the entire cycling plan and allow on-the-fly priority changes.

### How it works
All CNRs are generated and applied in a dormant/Pending state grouped by Spec.Priority.
Activation loop finds the next eligible priority: all lower priorities must be Successful and pass health checks.
Priority N is activated only when N-1 is healthy; activation proceeds level-by-level.

### Backward compatibility
Priority defaults to 0 for older specs or invalid values.
CNR activation remains driven by priority level

### DIFF
--- a/pkg/observer/priority.go
+++ b/pkg/observer/priority.go
@@ -72,53 +72,27 @@ func (c *controller) getPriority(nodeGroup *v1.NodeGroup) int32 {
 }
 
 // findNextActivatablePriority finds the lowest priority that can be activated
-func (c *controller) findNextActivatablePriority(allCNRs []v1.CycleNodeRequest) int32 {
+func (c *controller) findNextActivatablePriority(priorityGroups map[int32][]v1.CycleNodeRequest) int32 {
 	// Find the highest priority level that exists
 	maxPriority := int32(-1)
-	for _, cnr := range allCNRs {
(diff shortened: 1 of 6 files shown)


pkg/observer/priority.go
```diff
@@ -0,0 +1,431 @@
...
+	for _, nodeGroup := range changedNodeGroups {
+		priority := c.getPriority(nodeGroup.NodeGroup)
+		priorityGroups[priority] = append(priorityGroups[priority], nodeGroup)
+	}
+
+	return priorityGroups
+}
+
+// validatePriority validates that priority values are reasonable
+func (c *controller) validatePriority(priority int32) bool {
```
vincentportella, line 48: I'd argue we don't need to worry about a limit tbh, I think it was be any integer, just order them and apply

PR #118, opened 2025-08-11
check scaleUp has resulted in Ready nodes

MinyiZ:
Currently, ScaleUp is considered successful if `numReadyNodes = numHealthyASGNodes + numDetachedNodes`
This check was intentionally made relative rather than relying on an absolute expected node count in order to account for any autoscaling events.

However, this approach can cause issues when a newly launched node comes up unhealthy/unready and is then terminated by a node health-checking tool. In that case, the node disappears from both Kubernetes and the ASG, but Cyclops still considers the ScaleUp successful and proceeds to terminate nodes.

This PR adds an additional safeguard: verifying that at least one ready node has appeared after the ScaleUp start time. This ensures that ScaleUp is only marked as successful if at least one new, functioning node has actually joined the cluster.


pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -362,7 +363,8 @@ func (t *CycleNodeRequestTransitioner) transitionScalingUp() (reconcile.Result,
 		"scalingUpBy", len(t.cycleNodeRequest.Status.CurrentNodes))
 
 	// If either check isn't ready, requeue to check again later
-	if !(allInstancesReady && allKubernetesNodesReady) {
+	// also check if at least one Ready node was created after the scaleUpStarted time
+	if !(allInstancesReady && allKubernetesNodesReady && numNodesCreatedAfterScaleUpStarted > 0) {
```
mwhittington21, line 367: discuss: can we be more specific here to cover the edge case where mid-cycle for one node userdata change another userdata change is pushed out, and it causes all future nodes in the cycle to go unhealthy?

My initial suggestion for this is to have a different timestamp that is the time we last scaled up, not the time we initially entered the scaling up phase. Right now this timestamp is [set once](https://github.com/atlassian-labs/cyclops/blob/e2d1a0595f203a375e2ab1330c08a536046fdaae/pkg/controller/cyclenoderequest/transitioner/transitions.go#L302) when we transition from Initialised to ScalingUp.

If we track the last time we entered ScalingUp at all in a separate field, we would be able to ensure that for each batch of instances brought up that they are healthy and have been created.

pkg/controller/cyclenoderequest/transitioner/util_test.go
```diff
@@ -142,8 +144,52 @@ func TestFindOffendingNodes(t *testing.T) {
...
+		nodes         map[string]corev1.Node
+		cutoffTime    time.Time
+		expectedCount int
+	}{
+		{
+			name:          "empty nodes map",
+			nodes:         map[string]corev1.Node{},
+			cutoffTime:    baseTime,
+			expectedCount: 0,
+		},
```
mwhittington21, line 161: issue: missing test cases here for:

1. nodes exist but all after cutoff
2. nodes exist but all before cutoff

MinyiZ: thanks! added more test cases

pkg/controller/cyclenoderequest/transitioner/transitions.go
MinyiZ: I verified this by running it locally against a test cluster. The `ScaleUpStarted` is updated on every detach.
I ran on CNR on a nodegroup with 2 nodes with concurrency=1 and i see the time updated twice.

mwhittington21: Aha, excellent. Thank you for checking this. That sounds perfect then.

mwhittington21 (approved):
lgtm with the new test cases

PR #119, opened 2025-08-11
dnorman3/priority system

dnorman3:
### Summary
Implement priority-based creation in Cyclops. On each run, only the lowest-priority NodeGroups with detected changes get CNRs. When those CNRs complete successfully, the next run creates CNRs for the next priority level.

### Motivation
Reduce blast radius by ensuring lower-priority NodeGroups complete before higher ones are started.

### How it works
Observer finds changed NodeGroups and drops ones with in‑progress CNRs.
Compute the minimum NodeGroup.Spec.Priority among the remaining; create CNRs only for NodeGroups at that priority.
CNRs reconcile immediately via the existing CNR controller; no separate activation phase.
Subsequent runs repeat the process and naturally move to the next priority once lower levels are done.
If multiple NodeGroups share the same lowest priority, CNRs are created for all of them in that run.

### Backward compatibility
If NodeGroup.Spec.Priority is missing or negative, it’s treated as 0.
Priority is no longer stored on the CNR; ordering is decided solely from the NodeGroup when CNRs are created.
No changes required in the CNR controller.

### Notes
No grouping needed; a single pass to find the min priority and filter is sufficient.
Existing dropInProgressCNRs prevents overlapping work; dedup beyond that is optional.
Metrics remain unchanged (CNRsCreated per NodeGroup).

### DIFF
--- a/deploy/crds/atlassian.com_nodegroups_crd.yaml
+++ b/deploy/crds/atlassian.com_nodegroups_crd.yaml
@@ -308,8 +308,8 @@ spec:
                 type: array
               priority:
                 description: Priority controls the ordering of CNR creation for this
-                  NodeGroup. Lower values are higher priority. 0 runs first, then
-                  1, 2, ...
+                  NodeGroup. Lower values are higher priority. -10 runs before 0;
+                  then 10, 20, ...  Value range is from -2,147,483,648 to 2,147,483,647.
                 format: int32
                 type: integer
               skipInitialHealthChecks:
--- a/deploy/crds/atlassian.com_nodegroups_crd.yaml
+++ b/deploy/crds/atlassian.com_nodegroups_crd.yaml
@@ -306,6 +306,12 @@ spec:
                   - triggerEndpoint
                   type: object
                 type: array
+              priority:
+                description: Priority controls the ordering of CNR creation for this
+                  NodeGroup. Lower values are higher priority. 0 runs first, then
+                  1, 2, ...
+                format: int32
+                type: integer
               skipInitialHealthChecks:
                 description: SkipInitialHealthChecks is an optional flag to skip the
                   initial set of node health checks before cycling begins This does
--- a/pkg/apis/atlassian/v1/nodegroup_types.go
+++ b/pkg/apis/atlassian/v1/nodegroup_types.go
@@ -40,8 +40,8 @@ type NodeGroupSpec struct {
 	// SkipPreTerminationChecks is an optional flag to skip pre-termination checks during cycling
 	SkipPreTerminationChecks bool `json:"skipPreTerminationChecks,omitempty"`
 
-    // Priority controls the ordering of CNR creation for this NodeGroup.
-    // Lower values are higher priority. 0 runs first, then 1, 2, ...
+	// Priority controls the ordering of CNR creation for this NodeGroup.
+	// Lower values are higher priority. Examples: -10 runs before 0; then 10, 20, ...
     Priority int32 `json:"priority,omitempty"`
 }
 
--- a/pkg/apis/atlassian/v1/nodegroup_types.go
+++ b/pkg/apis/atlassian/v1/nodegroup_types.go
@@ -39,6 +39,10 @@ type NodeGroupSpec struct {
 
 	// SkipPreTerminationChecks is an optional flag to skip pre-termination checks during cycling
 	SkipPreTerminationChecks bool `json:"skipPreTerminationChecks,omitempty"`
+
+    // Priority controls the ordering of CNR creation for this NodeGroup.
+    // Lower values are higher priority. 0 runs first, then 1, 2, ...
+    Priority int32 `json:"priority,omitempty"`
 }
 
 // NodeGroupStatus defines the observed state of NodeGroup
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -97,7 +97,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-    return atlassianv1.CycleNodeRequest{
+	return atlassianv1.CycleNodeRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      finalName,
 			Namespace: namespace,
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -112,6 +112,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 			PreTerminationChecks:     nodeGroup.Spec.PreTerminationChecks,
 			SkipInitialHealthChecks:  nodeGroup.Spec.SkipInitialHealthChecks,
 			SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
+			Priority:                 nodeGroup.Spec.Priority,
 			ValidationOptions:        nodeGroup.Spec.ValidationOptions,
 		},
 	}
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -97,9 +97,6 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-    // Default/normalize priority for backward compatibility (not stored on CNR anymore)
-    _ = max(nodeGroup.Spec.Priority, 0)
-
     return atlassianv1.CycleNodeRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      finalName,
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -97,8 +97,8 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-    // Default/normalize priority for backward compatibility
-    priority := max(nodeGroup.Spec.Priority, 0)
+    // Default/normalize priority for backward compatibility (not stored on CNR anymore)
+    _ = max(nodeGroup.Spec.Priority, 0)
 
     return atlassianv1.CycleNodeRequest{
 		ObjectMeta: metav1.ObjectMeta{
@@ -114,8 +114,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 			HealthChecks:             nodeGroup.Spec.HealthChecks,
 			PreTerminationChecks:     nodeGroup.Spec.PreTerminationChecks,
 			SkipInitialHealthChecks:  nodeGroup.Spec.SkipInitialHealthChecks,
-			SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
-            Priority:                 priority,
+            SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
 			ValidationOptions:        nodeGroup.Spec.ValidationOptions,
 		},
 	}
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -97,7 +97,10 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-	return atlassianv1.CycleNodeRequest{
+    // Default/normalize priority for backward compatibility
+    priority := max(nodeGroup.Spec.Priority, 0)
+
+    return atlassianv1.CycleNodeRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      finalName,
 			Namespace: namespace,
@@ -112,7 +115,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 			PreTerminationChecks:     nodeGroup.Spec.PreTerminationChecks,
 			SkipInitialHealthChecks:  nodeGroup.Spec.SkipInitialHealthChecks,
 			SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
-			Priority:                 nodeGroup.Spec.Priority,
+            Priority:                 priority,
 			ValidationOptions:        nodeGroup.Spec.ValidationOptions,
 		},
 	}
--- a/pkg/generation/cnr.go
+++ b/pkg/generation/cnr.go
@@ -111,7 +111,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 			HealthChecks:             nodeGroup.Spec.HealthChecks,
 			PreTerminationChecks:     nodeGroup.Spec.PreTerminationChecks,
 			SkipInitialHealthChecks:  nodeGroup.Spec.SkipInitialHealthChecks,
-            SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
+			SkipPreTerminationChecks: nodeGroup.Spec.SkipPreTerminationChecks,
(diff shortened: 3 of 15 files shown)


pkg/generation/cnr.go
```diff
@@ -97,7 +97,10 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-	return atlassianv1.CycleNodeRequest{
+    // Default/normalize priority for backward compatibility (not stored on CNR anymore)
+    _ = max(nodeGroup.Spec.Priority, 0)
```
vincentportella, line 101: I don't follow what this is going

mwhittington21: Seems leftover, I agree this needs removal.

pkg/observer/controller.go
```diff
@@ -273,31 +273,77 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+        }
+    }
+    var filtered []*ListedNodeGroups
+    for _, g := range changedNodeGroups {
+        if getPriority(g.NodeGroup) == minPriority {
+            filtered = append(filtered, g)
+        }
+    }
+    return filtered
+}
```
vincentportella, line 324: use `changeNodeGroup` rather than `g`.

pkg/observer/controller.go
```diff
@@ -273,31 +273,77 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+            minPriority = p
+        }
+    }
+    var filtered []*ListedNodeGroups
+    for _, g := range changedNodeGroups {
+        if getPriority(g.NodeGroup) == minPriority {
+            filtered = append(filtered, g)
+        }
+    }
+    return filtered
```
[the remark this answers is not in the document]
dnorman3: done

pkg/observer/controller.go
dnorman3: done

pkg/generation/cnr.go
```diff
@@ -97,7 +97,7 @@ func GenerateCNR(nodeGroup atlassianv1.NodeGroup, nodes []string, name, namespac
 		}
 	}
 
-	return atlassianv1.CycleNodeRequest{
+    return atlassianv1.CycleNodeRequest{
```
mwhittington21, line 100: nit: due to what I assume is IDE settings we're currently mixing tabs with spaces. Please ensure your IDE is configured for tabs or fix the linter to pick this up and fix it, so we don't get a mix in the files.

pkg/observer/controller.go
```diff
@@ -273,31 +273,76 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
-			if c.DryMode {
-				drymodeStr = "[drymode] "
-			}
-			klog.V(2).Infof("%ssuccessfully applied cnr %q for nodegroup %q", drymodeStr, name, nodeGroup.NodeGroup.Name)
-			c.CNRsCreated.WithLabelValues(nodeGroup.NodeGroup.Name).Inc()
-		}
-	}
+// isLowerPriorityInProgress returns true if any in-progress CNR belongs to a NodeGroup with
+// a priority lower than the provided minPriority (i.e., must finish before creating higher priorities)
+func (c *controller) hasLowerPriorityCNRsInProgress(minPriority int32, inProg v1.CycleNodeRequestList) bool {
```
mwhittington21, line 327: nit: unless we really need to save space, prefer using full names for variables, so change this to `inProgress`

pkg/observer/controller.go
```diff
@@ -273,31 +273,58 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
 
 // createCNRs generates and applies CNRs from the changedNodeGroups
 func (c *controller) createCNRs(changedNodeGroups []*ListedNodeGroups) {
```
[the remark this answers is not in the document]
dnorman3: i have updated the code to do this

pkg/observer/controller.go
```diff
@@ -273,31 +273,76 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+// isLowerPriorityInProgress returns true if any in-progress CNR belongs to a NodeGroup with
+// a priority lower than the provided minPriority (i.e., must finish before creating higher priorities)
+func (c *controller) hasLowerPriorityCNRsInProgress(minPriority int32, inProg v1.CycleNodeRequestList) bool {
+    if len(inProg.Items) == 0 {
+        return false
+    }
+    // Build a list of valid nodegroups to map CNRs to priorities
+    allNG := c.validNodeGroups()
+    for _, cnr := range inProg.Items {
+        for i := range allNG.Items {
```
mwhittington21, line 334: Why are we mixing index-based loops with item assignation loops? Change this index based loop and the line below to be one line: `for _, ng := range allNG.Items {`

pkg/observer/controller.go
```diff
@@ -334,12 +380,31 @@ func (c *controller) Run() {
...
 	klog.V(3).Infof("waiting for %v to allow changes to settle", c.WaitInterval)
 	select {
-	case <-time.After(c.WaitInterval):
-		klog.V(3).Infof("applying %d CNRs", len(changedNodeGroups))
-		c.createCNRs(changedNodeGroups)
+    case <-time.After(c.WaitInterval):
+        klog.V(3).Infof("applying %d CNRs (lowest priority batch)", len(lowestPriorityBatch))
+        c.ApplyingNodeGroups.WithLabelValues().Set(float64(len(lowestPriorityBatch)))
+        c.createCNRs(lowestPriorityBatch)
+        c.ApplyingNodeGroups.WithLabelValues().Set(0)
```
mwhittington21, line 407: issue: setting this immediately back to 0 will overwrite the metric you set above if we are scraping via prometheus, as the metric is only captured every scrape interval (e.g. 15s or 30s). I think the way to handle this situation is to make the metrics counters, and increment them by the number that we applied. Then we can use `rate()` functions to view the changes.

pkg/observer/controller_test.go
```diff
@@ -387,3 +389,378 @@ func Test_dropInProgressNodeGroups(t *testing.T) {
...
+    _ = ctrl.client.Update(context.TODO(), &cnrA)
+
+    // Subsequent runs should never create B while A is Failed (treated as in-progress)
+    ctrl.Run()
+    ctrl.Run()
+    lst2, _ := generation.ListCNRs(ctrl.client, &client.ListOptions{Namespace: ctrl.Namespace})
+    assert.Equal(t, 1, len(lst2.Items))
+}
+
+func TestPriority_DefaultingNegativeAsZero_Sequential(t *testing.T) {
```
mwhittington21, line 501: Agree with @vincentportella from other comments on this PR that we should just allow negative numbers, but we should still default to 0 if not provided. That would make this test redundant.

pkg/observer/controller_test.go
```diff
@@ -387,3 +389,378 @@ func Test_dropInProgressNodeGroups(t *testing.T) {
...
+    _ = ctrl.client.Get(context.TODO(), client.ObjectKey{Namespace: ctrl.Namespace, Name: lst1.Items[0].Name}, &cnrA)
+    cnrA.Status.Phase = atlassianv1.CycleNodeRequestSuccessful
+    _ = ctrl.client.Update(context.TODO(), &cnrA)
+
+    ctrl.Run()
+    lst2, _ := generation.ListCNRs(ctrl.client, &client.ListOptions{Namespace: ctrl.Namespace})
+    assert.Equal(t, 2, len(lst2.Items))
+}
+
+func TestPriority_ThreeLevels_Sequential(t *testing.T) {
```
mwhittington21, line 535: I'd like to see one more complex test case, which is that if changes keep coming in for a lower priority node group then Cyclops will keep putting off the higher priority node group from cycling.

You can reuse the scaffolding from this test to start it, and make it do something like this:
1. A is created
2. A succeeds
3. B is created
4. New changes come in for A
5. A is created
6. B succeeds
7. C is NOT created yet
8. A succeeds
9. C is now created
10. C succeeds

mwhittington21 (changes requested):

pkg/observer/controller.go
```diff
@@ -273,31 +273,76 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
-		}
-	}
+// hasLowerPriorityCNRsInProgress returns true if any in-progress CNR belongs to a NodeGroup with
+// a priority lower than the provided minPriority (i.e., must finish before creating higher priorities)
+func (c *controller) hasLowerPriorityCNRsInProgress(minPriority int32, inProg v1.CycleNodeRequestList) bool {
+    if len(inProg.Items) == 0 {
+        return false
+    }
+    // Build a list of valid nodegroups to map CNRs to priorities
+    allNG := c.validNodeGroups()
```
vincentportella, line 332: Follow-up from @mwhittington21 comment regarding `inProg`, that should apply to all variables here. So this should become `allNodegroups`.

pkg/observer/controller.go
```diff
@@ -273,31 +273,75 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+    }
+}
 
-		name := generation.GetName(cnr.ObjectMeta)
+// selectLowestPriorityNodeGroups returns only the node groups at the lowest priority value
+func (c *controller) selectLowestPriorityNodeGroups(changedNodeGroups []*ListedNodeGroups) []*ListedNodeGroups {
+    if len(changedNodeGroups) == 0 {
+        return nil
+    }
+    getPriority := func(ng *v1.NodeGroup) int32 { return max(ng.Spec.Priority, 0) }
```
vincentportella, line 308: We don't need the function, the default value is 0 so if it isn't set in kube, it is still safe to use. Also this is still implying that the minimum value will be 0.

mwhittington21: @dnorman3 and I chatted about backwards compatibility of the function. I'm not too concerned, but he pointed out that because the value will default to 0, in a cluster with some nodegroups with no value set and some with values set, the ones with no values set would get set to 0.

There are two options from there:
1. What @dnorman3 has done here, plus adding validation on the CRD itself to prevent <0 values. I think this is a fine solution, the only downside being that it involves slightly more code and a harder rollout procedure.
2. What @vincentportella is saying, which is we just let them default to 0 and if anyone has mixed nodegroups with negatives those will get negative priority. The edge case only affects clusters that are using a mix of node groups with defaulted priority, and node groups with manually specified negative priority. This is also fine.

After writing it out, I think we should go for number 2 because it simplifies the rollout and doesn't involve changing the CRD spec. I don't think the edge case will affect many customers - most will probably never specify priority.

pkg/observer/controller.go
```diff
@@ -273,31 +273,75 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+    }
+}
 
-		name := generation.GetName(cnr.ObjectMeta)
+// selectLowestPriorityNodeGroups returns only the node groups at the lowest priority value
+func (c *controller) selectLowestPriorityNodeGroups(changedNodeGroups []*ListedNodeGroups) []*ListedNodeGroups {
+    if len(changedNodeGroups) == 0 {
+        return nil
+    }
+    getPriority := func(ng *v1.NodeGroup) int32 { return ng.Spec.Priority }
```
mwhittington21, line 308: nit(minor optimisation): getPriority doesn't need to be a function, it can just be a parameter access.

pkg/apis/atlassian/v1/nodegroup_types.go
```diff
@@ -39,6 +39,10 @@ type NodeGroupSpec struct {
 
 	// SkipPreTerminationChecks is an optional flag to skip pre-termination checks during cycling
 	SkipPreTerminationChecks bool `json:"skipPreTerminationChecks,omitempty"`
+
+    // Priority controls the ordering of CNR creation for this NodeGroup.
+    // Lower values are higher priority. 0 runs first, then 1, 2, ...
```
mwhittington21, line 44: nit: fix comment to include example negative number, or reference negative numbers are possible

pkg/observer/metrics.go
```diff
@@ -53,6 +57,32 @@ func newMetrics() *metrics {
...
+        ),
+        ApplyingNodeGroups: prometheus.NewCounterVec(
+            prometheus.CounterOpts{
+                Name:      "applying_nodegroups",
+                Namespace: metricsNamespace,
+                Help:      "counter of nodegroups applied per run (lowest priority batch)",
+            },
+            []string{},
+        ),
+        BlockedNodeGroups: prometheus.NewCounterVec(
```
mwhittington21, line 78: note: I'm not entirely sure how well these metrics will work but I guess we'll see once we launch the change. Better to have them than not, but I'm sure we can make tweaks to their usefulness later.

mwhittington21 (changes requested):

deploy/crds/atlassian.com_nodegroups_crd.yaml
```diff
@@ -306,6 +306,12 @@ spec:
                   - triggerEndpoint
                   type: object
                 type: array
+              priority:
+                description: Priority controls the ordering of CNR creation for this
+                  NodeGroup. Lower values are higher priority. 0 runs first, then
```
vincentportella, line 311: Similar to @mwhittington21 suggestion for the comment, this should just say that any number is usable

pkg/observer/controller.go
```diff
@@ -334,12 +381,31 @@ func (c *controller) Run() {
 		}
 	}
 
-	// wait for the desired amount to allow any in progress changes to batch up
+    // Filter to only the lowest priority nodegroups
+    lowestPriorityBatch := c.selectLowestPriorityNodeGroups(changedNodeGroups)
+    if len(lowestPriorityBatch) == 0 {
```
vincentportella, line 386: nit: doesn't this imply that there are no nodegroups to apply at all? Otherwise you would always get at least 1 nodegroup in the lowest priority batch

pkg/observer/metrics.go
vincentportella: @dnorman3 could you run this in your lab and write a page showing what the metrics look like so we can see before merging this change?

mwhittington21:
LGTM just waiting on the output of the lab-based tests

pkg/observer/controller.go
```diff
@@ -273,31 +273,74 @@ func (c *controller) dropInProgressNodeGroups(nodeGroups v1.NodeGroupList, cnrs
...
+            klog.V(2).Infof("%ssuccessfully applied cnr %q for nodegroup %q", drymodeStr, name, nodeGroup.NodeGroup.Name)
+            c.CNRsCreated.WithLabelValues(nodeGroup.NodeGroup.Name).Inc()
+        }
+    }
+}
 
-		name := generation.GetName(cnr.ObjectMeta)
+// selectLowestPriorityNodeGroups returns only the node groups at the lowest priority value
+func (c *controller) selectLowestPriorityNodeGroups(changedNodeGroups []*ListedNodeGroups) []*ListedNodeGroups {
+    if len(changedNodeGroups) == 0 {
```
vincentportella, line 305: this is a redundant check because `changedNodeGroups` will never be `0` here since you already check there are some prior to running this function, and if `nil` were ever to be returned here anyway that would currently crash observer because we check the first element of a `nil` list right after the call of this function.

mwhittington21: Yes please create and return an empty list if you end up keeping this check.

mwhittington21 (approved):

vincentportella (dismissed):
lgtm

mwhittington21 (approved):

vincentportella (approved):

PR #134, opened 2026-01-27
Coordinate Cyclops with Cluster Autoscaler to prevent node upgrade conflicts

dnorman3:
# Cluster Autoscaler Annotation Management - Feature Summary

Adds annotation management to coordinate between Cyclops and Cluster Autoscaler during node cycling, preventing Cluster Autoscaler from removing new nodes before old nodes are drained.

## Problem

During node cycling, Cluster Autoscaler can remove new nodes before Cyclops finishes draining old nodes, causing cycling to fail or nodes to be terminated prematurely.

## Solution

- Adds `cluster-autoscaler.kubernetes.io/scale-down-disabled: "true"` annotation to new nodes during ScalingUp phase
- Removes annotation in both Successful and Healing phase transitions
- **Marker annotation tracking:** Uses `cyclops.atlassian.com/annotation-managed` to track ownership (distinguishes Cyclops-managed vs pre-existing annotations)
- **Pre-existing preservation:** Preserves annotations set by ASG Launch Templates or other external sources
- **Opt-out feature:** NodeGroup annotation `cyclops.atlassian.com/disable-annotation-management: "true"` disables management (default: enabled)

## Key Changes

**pkg/controller/cyclenoderequest/transitioner/util.go:**
- `addScaleDownDisabledAnnotation()` - Adds annotation + marker (preserves pre-existing)
- `cleanupScaleDownDisabledAnnotations()` - Removes annotations only if marker present
- `shouldManageAnnotations()` - Checks NodeGroup opt-out annotation
- `getNodeGroup()` - Finds matching NodeGroup resource

**pkg/controller/cyclenoderequest/transitioner/transitions.go:**
- `transitionScalingUp()` - Adds annotations (if enabled)
- `transitionSuccessful()` / `transitionHealing()` - Removes annotations (if added)

**pkg/controller/cyclenoderequest/transitioner/transitioner.go:**
- Constants: `nodeGroupAnnotationKey`, `cyclopsManagedAnnotation`

## Opt-Out Configuration

```yaml
apiVersion: atlassian.com/v1
kind: NodeGroup
metadata:
  annotations:
    cyclops.atlassian.com/disable-annotation-management: "true"  # Disables management
```

**Why NodeGroup?** Persistent configuration (survives CNR deletion), aligned with node lifecycle, prevents conflicts with overlapping CNRs.

## Design Decisions

- Marker annotation ensures safe cleanup (only removes Cyclops-managed annotations)
- Pre-existing annotations preserved for backward compatibility
- Best-effort operations (don't block cycling)
- Dual cleanup paths (Successful + Healing phases)
- Opt-out approach (default enabled)

## Verification

```bash
# Check nodes with annotations
kubectl get nodes -o json | jq -r '.items[] | select(.metadata.annotations["cluster-autoscaler.kubernetes.io/scale-down-disabled"] == "true") | .metadata.name'

# Check opt-out configuration
kubectl get nodegroup <name> -o jsonpath='{.metadata.annotations.cyclops\.atlassian\.com/disable-annotation-management}'

# Enable opt-out
kubectl annotate nodegroup <name> cyclops.atlassian.com/disable-annotation-management="true"
```


mwhittington21:
@dnorman3 I think you also need to add logic to remove this label from a node if we transition to healing. The logic would go somewhere [here](https://github.com/atlassian-labs/cyclops/blob/master/pkg/controller/cyclenoderequest/transitioner/transitions.go#L536). The reason for this is, if a CNR fails we need to remove the label or the node will forever be ignored by cluster-autoscaler.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -492,6 +513,14 @@ func (t *CycleNodeRequestTransitioner) transitionCordoning() (reconcile.Result,
 func (t *CycleNodeRequestTransitioner) transitionWaitingTermination() (reconcile.Result, error) {
 	t.rm.LogEvent(t.cycleNodeRequest, "WaitingTermination", "Waiting for instances to terminate")
 
+	// Periodically attempt to cleanup scale-down-disabled annotations in case previous cleanup attempts failed.
+	// This ensures annotations don't remain on nodes permanently if cleanup failed earlier.
+	// This is idempotent and safe to call multiple times.
+	if err := t.cleanupScaleDownDisabledAnnotations(); err != nil {
```
mwhittington21: I think rather than cleaning up the annotation at large like this, it should be focused on nodes we've been working on. Adding it to the [healing logic](https://github.com/atlassian-labs/cyclops/blob/master/pkg/controller/cyclenoderequest/transitioner/transitions.go#L536) should accomplish this.

I worry that by doing something like this, we could be removing labels that we or customers have added themselves. Or other pieces of software might have added it too.

mwhittington21 (changes requested):

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -510,6 +530,14 @@ func (t *CycleNodeRequestTransitioner) transitionWaitingTermination() (reconcile
 
 // transitionHealing handles healing CycleNodeRequests
 func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, error) {
+	// Clean up scale-down-disabled annotations from nodes we annotated during this cycling operation.
+	// This ensures annotations don't remain if cycling fails and we need to rollback.
+	// This is best-effort and failures should not block the healing process.
+	if err := t.cleanupScaleDownDisabledAnnotations(); err != nil {
```
mwhittington21: I would prefer it if the cleanup wasn't done in such a wide-reaching way here, but instead just targets the nodes similar to how the reattachment code below works. There's a for loop over the nodes, if we remove the labels from nodes one by one during that loop it would allow the logic to mirror how other cleanup works.

mwhittington21 (dismissed):
lgtm

mwhittington21 (dismissed):

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -584,6 +580,11 @@ func (t *CycleNodeRequestTransitioner) transitionHealing() (reconcile.Result, er
 		}
 	}
 
+	// remove the scale-down-disabled annotation from any new nodes that were created during this cycle
+	// This is best-effort and failures should not block the healing process
+	// Called outside the loop to ensure it runs even if NodesToTerminate is empty (e.g., all nodes already terminated)
+	t.cleanupScaleDownDisabledAnnotations()
```
dnorman3: Previously, cleanupScaleDownDisabledAnnotations() was called inside the loop over NodesToTerminate in transitionHealing(). If NodesToTerminate was empty (e.g., all nodes already terminated) or all nodes didn't exist, the loop didn't run, so cleanup never executed. This left annotations on new nodes created during ScalingUp, preventing Cluster Autoscaler from managing them.

Moved the cleanup call outside the loop so it always runs in the Healing phase, regardless of NodesToTerminate state. The cleanup finds all nodes created after ScaleUpStarted (by creation timestamp), not just nodes in NodesToTerminate, so it correctly targets the new replacement nodes that have the annotation.

Additionally, changed cleanupScaleDownDisabledAnnotations() to use listNodes() instead of listReadyNodes() so it processes nodes even if they're not Ready yet. This ensures cleanup happens when cycling fails in Healing, even if replacement nodes are still initialising. The annotation prevents Cluster Autoscaler from scaling down nodes, so removing it during failure recovery is important regardless of node readiness state.

pkg/controller/cyclenoderequest/transitioner/transitioner.go
```diff
@@ -17,6 +17,14 @@ type transitionFunc func() (reconcile.Result, error)
 // different request types.
 const cycleNodeLabel = "cyclops.atlassian.com/terminate"
 
+// clusterAutoscalerScaleDownDisabledAnnotation is the annotation key used to prevent
+// Cluster Autoscaler from scaling down a node. This is used to protect new nodes
+// during the cycling process from being removed by Cluster Autoscaler before
+// the corresponding old nodes are fully terminated.
+// See: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-prevent-cluster-autoscaler-from-scaling-down-a-particular-node
+const clusterAutoscalerScaleDownDisabledAnnotation = "cluster-autoscaler.kubernetes.io/scale-down-disabled"
```
vincentportella, line 30: could we generalise this behaviour and give the annotation via the CNR instead?

dnorman3: A few reasons why i didn't

The annotation is tied to a node lifecycle, not a cnr lifecycle. If annotations were the in cnr spec, deleting a cnr before cleanup could leave stale annotations. 

Autoscaler reads annotations from a node object, not the cnr . If we put them in cnr spec, we would have  to sync them to nodes.

cnrs  can affect overlapping nodegroups, having annotations in cnr spec can create race conditions and ownership issues.

vincentportella: > If we put them in cnr spec, we would have to sync them to nodes.

This is exactly what I'm suggesting. We are already doing this in the code, I am just proposing that the annotation(s) get defined in the CNR rather than hardcoded.

> cnrs can affect overlapping nodegroups

Each CNR is generated by observer for a single nodegroup. If a CNR affects nodes across nodegroups that would be a misconfiguration somewhere.

mwhittington21: I don't think we need to be generic here, the annotations are very specific to cluster-autoscaler. If we need another label, we probably want a code change because it will need to be used in specific conditions.

mwhittington21: What this comment has caused me to re-evaluate is that we should add a field on the CNR to enable/disable this behaviour and default it to true. That way if any users of the software have an issue with this particular annotation/workflow, they can disable the feature and keep their cluster working.

vincentportella: If we have the annotation defined in the CNR then that achieve's it. It wouldn't be enabled by default because the annotations need to be added for the behaviour to take effect. I'd argue this feature is just about adding annotations to new instances and then removing them later. We are doing it for cluster-autoscaler of course but it can apply more widely.

dnorman3: i added the option to opt out of it in the nodegroup (not cnr) see https://github.com/atlassian-labs/cyclops/commit/67d6d0f1f25f5c727f7fb463c8392bcebde1ff98#commitcomment-175885640 for reasoning

mwhittington21: I like putting it on the nodegroup, it feels like something that should be associated with a base nodegroup config.

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -613,6 +613,41 @@ func (t *CycleNodeRequestTransitioner) getNodegroupForMetrics() string {
...
+		if t.cycleNodeRequest.IsFromNodeGroup(*ng) {
+			return ng, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// shouldManageAnnotations returns true if annotations should be managed.
+func (t *CycleNodeRequestTransitioner) shouldManageAnnotations() bool {
```
mwhittington21, line 643: @dnorman3 please add tests for this to ensure that the overall process doesn't add the annotation to nodes if the nodegroup has this annotation present and with the right value.

mwhittington21 (approved):

Jacobious52 (approved):

PR #135, opened 2026-02-03
add retry logic

castlemilk:
adds retry login for intermiitent aws errors, where this would result in cyclops putting a CNR into `Failed` state, when getting error from AWS like:

```
Post "https://autoscaling.us-west-2.amazonaws.com/": dial tcp 18.246.117.35:443: i/o timeout, RequestError: send request failed
```


pkg/cloudprovider/aws/builder.go
```diff
@@ -2,26 +2,84 @@ package aws
...
+	httpClient := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+
+	config := &aws.Config{
+		Credentials: creds,
+		// Maximum number of retries for API calls
+		MaxRetries: aws.Int(5),
+		// Custom retryer with exponential backoff
+		Retryer: customRetryer{
```
mwhittington21: The golang SDK [docs](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-retries-timeouts.html) suggest that we should just use a standard Retryer and not worry about all the custom logic. Have you looked into whether this would be sufficient?

Specifically https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/retry#Standard

mwhittington21 (changes requested):

pkg/cloudprovider/aws/builder.go
acastlemilk: done! I had planned to do both, but can start with this first

acastlemilk (commented):
done

mwhittington21 (approved):

aqasimi (approved):

dtnyn (approved):

PR #140, opened 2026-05-07
aws: tag EC2 instance with detach timestamp on DetachInstance

priyeshgpatel:
When a CycleNodeRequest fails mid-cycle and the healing path is unable to re-attach the detached instance (e.g. due to missing IAM permissions or a transient error), the instance continues running outside the ASG — invisible to the autoscaler, not terminated, and accumulating cost.

Tag the instance with `cyclops:detached-at=<RFC3339 timestamp>` immediately after a successful DetachInstances call. This makes orphaned instances trivially discoverable:

  aws ec2 describe-instances \
    --filters "Name=tag-key,Values=cyclops:detached-at"

Tag failure is non-fatal — logged but does not affect the detach result. Also add a no-op CreateTags implementation to the fake EC2 client so tests continue to pass.

  ---
  IAM permissions required

  To use this feature, add ec2:CreateTags to the cyclops-operator IAM role policy, scoped to cluster-owned instances and restricted to the cyclops:detached-at tag key only:

  {
    "Sid": "EC2TagPermission",
    "Effect": "Allow",
    "Action": "ec2:CreateTags",
    "Resource": "*",
    "Condition": {
      "StringEquals": {
        "ec2:ResourceTag/kubernetes.io/cluster/<cluster-name>": "owned"
      },
      "ForAllValues:StringEquals": {
        "aws:TagKeys": ["cyclops:detached-at"]
      }
    }
  }

  Tag failure is non-fatal — if the permission is missing, cyclops logs the error and continues normally.

### DIFF
--- a/pkg/cloudprovider/aws/aws.go
+++ b/pkg/cloudprovider/aws/aws.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/atlassian-labs/cyclops/pkg/cloudprovider"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


mwhittington21:
@priyeshgpatel hi, thanks for the contribution! The change looks simple, useful, and well contained. It should have an associated test for the behaviour so it is ensured to remain working. If you add tests it should be good to go.

PR #141, opened 2026-05-13
upgrade from operator-sdk to controller-gen for generating CRDs to use more up to date tooling

mwhittington21:
I am changing the CNR CRD for the first time in a long time and found it easier to use up to date tooling rather than the antique version of operator-sdk we were using to be able to generate our CRDs.

Kept the file names the same in case anyone is applying them directly from the repo.


awprice (commented):
Conceptually a great change — swapping the ancient operator-sdk v0.19.0 build-from-source workflow for controller-gen is the right move. A few things worth addressing before merge:

**[Blocking] 1. Stale docs reference**
`docs/README.md:93` still points readers to `make install-operator-sdk` which no longer exists after this PR. The developer instructions will be broken. Worth updating to reference `controller-gen` and the `go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0` one-liner. `README.md:10` also mentions operator-sdk in the project description — a minor tweak there would be good too.

**[Blocking] 2. `zz_generated.deepcopy.go` not regenerated**
The PR adds a `generate-deepcopy` target (`controller-gen object …`) but the regenerated file isn't included in this PR. The existing file still has `// Code generated by operator-sdk. DO NOT EDIT.` in its header. Worth running `make generate-deepcopy` and committing the result here to confirm controller-gen produces an equivalent file and to avoid the next contributor getting a noisy unrelated diff.

**[Blocking] 3. Pin the controller-gen version**
The Makefile comment mentions `v0.14.0` but nothing enforces it. If a contributor has a different version installed they'll silently generate a different `controller-gen.kubebuilder.io/version` annotation and cause CRD churn in future PRs. The kubebuilder canonical pattern is a `tools.go` + `go.mod` tools dependency with a `bin/controller-gen` make target that builds the pinned version, then references `$(CONTROLLER_GEN)` everywhere. At minimum, using `go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0` directly in the Makefile would make it hermetic.

**[Nice-to-have] 4. `zz_generated.defaults.go` and `zz_generated.openapi.go` are now orphaned**
controller-gen doesn't produce these files and they're no longer part of any documented generation flow. Worth deciding whether to delete them or document what generates them — leaving them as-is indefinitely will rot.

dnorman3 (dismissed):

awprice (approved):

MinyiZ (approved):

PR #142, opened 2026-05-13
Fix various situations where cluster autoscaler scale-down-disabled annotation was removed early, or left on nodes forever

mwhittington21:
Merge after https://github.com/atlassian-labs/cyclops/pull/141 which separates out a change to how the CRDs are generated to make the diff easier to digest here.

The `cluster-autoscaler.kubernetes.io/scale-down-disabled` annotation is added to new nodes during cycling to ensure cluster-autoscaler does not terminate them if pods are taking a long time to drain onto the new nodes. There were some bugs found with the implementation that led to undesirable situations.

**Bug 1:** Multi-batch annotation cleanup misses earlier batches. With `concurrency < total nodes`, ScaleUpStarted is overwritten each batch. When cleanup runs at Successful, only the last batch's replacement nodes are found — earlier batches' scale-down-disabled annotations are left permanently. **When concurrency == 1, this leaves a large amount of nodes with this annotation, increasing running costs of the cluster significantly.**

**Bug 2:** Old Successful CNRs strip annotations from active CNRs on pod restart. After a restart, all CNRs are re-queued. Old Successful CNRs call cleanupScaleDownDisabledAnnotations() with stale ScaleUpStarted timestamps, removing annotations from nodes actively protected by newer CNRs. **This exposes replacement nodes to cluster autoscaler scale-down mid-cycle.**

This PR fixes the logic paths that led to these bugs, but also adds a new "node controller" which catches any future slips in logic. Any nodes which were annotated by the process but are no longer targeted by an active CNR will have the annotations removed.

The fix for both bugs involves keeping track of when the annotations are added and removed by adding a new list to the CNR object. This avoids the need for discovering nodes using time-based methods.

I also optimised the logic of annotation management to happen when each batch of concurrency is processed, rather than removing annotations at the end of the entire CNR when it transitions to a terminal state. This allows nodes that were rotated to be scaled down by cluster-autoscaler if demand for them has disappeared, while still avoiding termination of the new nodes while they're still empty.


pkg/metrics/metrics.go
```diff
@@ -131,6 +131,32 @@ var (
 		},
 		[]string{"nodegroup"},
 	)
+
+	// --- Node Cleanup Controller metrics (safety-net reconciler) ---
```
awprice: Both of these metrics mention a "Node Cleanup controller" - but the controller is just called "Node controller" - should the metric names be updated?

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// NewReconciler returns a new Reconciler for Nodes and registers it with the manager.
+// It reuses the manager's shared Node cache (already populated by the CNR transitioner)
+// to avoid duplicating API traffic.
+func NewReconciler(mgr manager.Manager, namespace string) (reconcile.Reconciler, error) {
+	rawClient := kubernetes.NewForConfigOrDie(mgr.GetConfig())
+
+	reconciler := &Reconciler{
+		mgr:       mgr,
```
awprice: Is mgr even accessed outside of NewReconciler? Wondering if we can get rid of it

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	// clusterAutoscalerScaleDownDisabledAnnotation is the annotation key used to prevent
+	// Cluster Autoscaler from scaling down a node.
+	clusterAutoscalerScaleDownDisabledAnnotation = "cluster-autoscaler.kubernetes.io/scale-down-disabled"
+)
+
+var log = logf.Log.WithName(controllerName)
+
+// Reconciler watches Node objects and cleans up stale Cyclops-managed annotations
+// left behind when a CycleNodeRequest is deleted before reaching a terminal phase.
+type Reconciler struct {
```
awprice, line 57: This probably doesn't need to be public

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	controllerName       = "node.controller"
+	reconcileConcurrency = 1
+	requeueAfter         = 5 * time.Minute
```
awprice: Should we make both reconcileConcurrency and requeueAfter configurable and passed into NewReconciler?

awprice: Maybe even the annotations too?

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+
+func hasAnnotation(annotations map[string]string, key string) bool {
+	_, ok := annotations[key]
+	return ok
+}
+
+// hasCyclopsManagedAnnotation is a controller-runtime predicate that passes only
+// Node events where the node carries the cyclopsManagedAnnotation. This avoids
+// reconciling every node in the cluster.
+type hasCyclopsManagedAnnotation struct{}
```
awprice: nit - Maybe this predicate could be a re-usable predicate for if an object has an annotation, put in a util package?

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	if active {
+		logger.V(1).Info("Node is still involved in an active CycleNodeRequest, requeueing", "requeueAfter", requeueAfter)
+		metrics.NodeCleanupReconciles.WithLabelValues("active_cnr_skipped").Inc()
+		return reconcile.Result{RequeueAfter: requeueAfter}, nil
+	}
+
+	// No active CNR covers this node — clean up the stale annotations.
+	logger.Info("Removing stale scale-down-disabled annotations from node")
+
+	if err := removeAnnotationWithRetry(node.Name, clusterAutoscalerScaleDownDisabledAnnotation, r.rawClient); err != nil {
```
awprice: I notice we're removing each annotation individually - what happens if one is removed successfully and not the other?

Should we be removing them atomically? i.e. both at the same time in the one api call?

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,256 @@
...
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// isTerminalPhase returns true for phases where the CNR lifecycle has ended and
+// annotation cleanup has already been handled (or is no longer expected).
+func isTerminalPhase(phase atlassianv1.CycleNodeRequestPhase) bool {
```
awprice: I think this should be a function on the CNR object or the CNR status as we are in theory spreading decisions around what is considered a terminal status in more places

pkg/controller/node/controller.go
awprice: We also have removeScaleDownDisabledAnnotation in util.go - and now this new logic to remove annotations - should we move it all into a single location/function?

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -532,6 +532,14 @@ func (t *CycleNodeRequestTransitioner) transitionWaitingTermination() (reconcile
 		return t.transitionToHealing(err)
 	}
 
+	// When a batch completes and we're looping back to Initialised, clean up
+	// the annotations from this batch's replacement nodes. Their old counterparts
+	// have been terminated, so the protection is no longer needed. This avoids
+	// accumulating protected nodes across the entire cycle duration.
+	if desiredPhase == v1.CycleNodeRequestInitialised && t.shouldManageAnnotations() {
+		t.cleanupScaleDownDisabledAnnotations()
```
awprice, line 546: Should our new node controller be doing this?

On second thoughts this needs access to the CNR to update annotated nodes so maybe not

awprice: I guess our node controller is really just to catch edge cases and we're still relying on the CNR transitioner to do annotation removal on the good path

docs/node-cleanup-reconciler.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Tests
+
+All tests are in `pkg/controller/node/controller_test.go`. Run with:
+
+```
+go test ./pkg/controller/node/ -v
+```
+
+**Unit tests (10 cases):** Table-driven, covering:
```
awprice: Listing out the tests here is probably redundant - I could see this not being kept up to date over time

awprice (changes requested):

pkg/controller/cyclenoderequest/transitioner/transitions.go
mwhittington21: Correct, I made the node controller first and tested it to act as a backstop. I have plans to eventually put more logic into it to clean up stuck nodes/finalizers/etc, but I wanted to start simple.

The node controller actually doesn't catch anything currently since I fixed the bugs. I could leave it to the node controller, but that doesn't solve the issue of cost saving by allowing nodes that have already been acted on to scale down.

pkg/metrics/metrics.go
mwhittington21: I'll change the comments to reference "node controller" but leave the references to cleanup in the metric names.

pkg/controller/node/controller.go
mwhittington21: I'll push back on the annotations for now, but the others make sense.

pkg/controller/node/controller.go
mwhittington21: I'm not too concerned with removing each annotation individually. As long as the final cyclops-managed annotation goes last, the reconciler will eventually remove all other annotations. Pre-optimising is more likely to introduce a bug, so I plan to leave it for now.

pkg/controller/node/controller.go
mwhittington21: Gonna ignore this since it's following the patterns in the other 2 controllers and I can fix in a refactor PR later. I don't want to change a bunch of lines that are irrelevant to the feature just to make it easier to digest.

pkg/controller/node/controller.go
```diff
@@ -0,0 +1,221 @@
...
+	if active {
+		logger.V(1).Info("Node is still involved in an active CycleNodeRequest, requeueing", "requeueAfter", r.options.RequeueAfter)
+		metrics.NodeCleanupReconciles.WithLabelValues("active_cnr_skipped").Inc()
+		return reconcile.Result{RequeueAfter: r.options.RequeueAfter}, nil
+	}
+
+	// No active CNR covers this node — clean up the stale annotations.
+	logger.Info("Removing stale scale-down-disabled annotations from node")
+
+	if err := k8s.RemoveScaleDownDisabledAnnotationsFromNode(node.Name, r.rawClient); err != nil {
```
awprice, line 160: When the node controller removes annotations it goes through `k8s.RemoveScaleDownDisabledAnnotationsFromNode` which doesn't update the `NodesWithAnnotation` gauge. The transitioner's `removeScaleDownDisabledAnnotation` does set it to 0. So if the safety net fires, the gauge will be stale until something else resets it. Might want to add the gauge update here too.

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -40,6 +40,10 @@ func (t *CycleNodeRequestTransitioner) transitionToFailed(err error) (reconcile.
 // transitionToUnsuccessful transitions the current cycleNodeRequest to healing/failed
 func (t *CycleNodeRequestTransitioner) transitionToUnsuccessful(phase v1.CycleNodeRequestPhase, err error) (reconcile.Result, error) {
 	t.cycleNodeRequest.Status.Phase = phase
+
+	// Clear AnnotatedNodes so it is persisted as empty with this status update.
+	// This ensures subsequent requeues won't re-run cleanup.
+	t.cycleNodeRequest.Status.AnnotatedNodes = nil
```
awprice, line 50: The diff removes `cleanupScaleDownDisabledAnnotations()` from `transitionHealing` and instead nils out `AnnotatedNodes` in `transitionToUnsuccessful` (`util.go` line 46) without actually removing the annotations from the nodes. So if a CNR goes Healing → Failed, the annotations stay on the nodes and we lose the list of what to clean up. The node controller will eventually catch it but that's a 5 minute requeue cycle — is the trade-off here that we'd rather not slow down the failure path with cleanup? Might be worth a comment explaining why we deliberately skip cleanup here.

pkg/controller/cyclenoderequest/transitioner/transitions.go
```diff
@@ -532,6 +532,14 @@ func (t *CycleNodeRequestTransitioner) transitionWaitingTermination() (reconcile
 		return t.transitionToHealing(err)
 	}
 
+	// When a batch completes and we're looping back to Initialised, clean up
+	// the annotations from this batch's replacement nodes. Their old counterparts
+	// have been terminated, so the protection is no longer needed. This avoids
+	// accumulating protected nodes across the entire cycle duration.
+	if desiredPhase == v1.CycleNodeRequestInitialised && t.shouldManageAnnotations() {
```
awprice, line 545: When a batch completes and we loop back to Initialised, we immediately strip the scale-down-disabled annotation from this batch's replacement nodes. Those nodes just received all the pods drained from the old nodes though — if cluster-autoscaler decides they're underutilised and scales them down before the next batch's replacements come up, wouldn't that cause churn? Is there a reason we're not keeping the protection until the whole CNR completes?

pkg/controller/cyclenoderequest/transitioner/util.go
```diff
@@ -68,9 +72,21 @@ func (t *CycleNodeRequestTransitioner) transitionToUnsuccessful(phase v1.CycleNo
...
+		t.cleanupScaleDownDisabledAnnotations()
+	}
+
 	t.rm.LogEvent(t.cycleNodeRequest, "Successful", "Successfully cycled nodes")
 	t.cycleNodeRequest.Status.Phase = v1.CycleNodeRequestSuccessful
 
+	// Clear any remaining AnnotatedNodes (including failed-to-remove ones) so
+	// subsequent requeues don't re-run cleanup. The node controller will catch
+	// any that failed to be removed here.
+	t.cycleNodeRequest.Status.AnnotatedNodes = nil
```
awprice: `cleanupScaleDownDisabledAnnotations` (called on line 79) already sets `AnnotatedNodes` to just the `failedNodes` at the end — so if some removals fail, they're preserved for retry. But then line 88 unconditionally nils out `AnnotatedNodes` again, throwing away that failed list. Either the retry logic in `cleanupScaleDownDisabledAnnotations` keeping `failedNodes` is dead code, or this nil on line 88 is a bug. Which one is it?

pkg/metrics/metrics.go
```diff
@@ -154,6 +180,8 @@ func Register(client client.Client, logger logr.Logger, namespace string) {
 		NodesWithAnnotation,
 		AnnotationCleanupAttempts,
```
awprice: nit - `AnnotationCleanupAttempts` and `AnnotationCleanupDuration` are registered but nothing ever increments them. Were you planning to wire these up in `cleanupScaleDownDisabledAnnotations`, or should we just remove them?

awprice (changes requested):

pkg/metrics/metrics.go
mwhittington21: This is a good find, not introduced by this PR but something I'll fix up while I'm here. Thanks.

awprice (dismissed):
LGTM!

MinyiZ (dismissed):

MinyiZ (approved):

awprice (approved):

PR #145, opened 2026-05-19
further prep for integration tests

awprice:
# Further prep for integration tests

Follow-up to #144. Two production changes and their unit tests.

### `pkg/k8s` — use merge patch for node annotations and labels

`AddAnnotationToNode` and `AddLabelToNode` previously used a JSON Patch
`"add"` operation. This fails at the apiserver when the node's
`metadata.annotations` / `metadata.labels` map is nil — something that
happens in envtest (no kubelet to initialise the map) and occasionally in
real clusters before a kubelet has written its own annotations.

Both functions now use a strategic merge patch via the new `MergePatchNode`
helper, which the apiserver handles correctly regardless of whether the map
exists. Behaviour in normal clusters is identical.

8 unit tests added in `pkg/k8s/node_test.go`, including an explicit
nil-annotations case that would have failed before this change.

### `pkg/manager` — extract controller wiring into `Run`

`cmd/manager/main.go` used to inline scheme registration and all three
reconciler registrations. The integration test `world` package had to
duplicate that wiring, meaning any future change to controller setup could
silently diverge between production and tests.

`pkg/manager.Run(ctx, mgr, deps)` is now the single place that registers
the cyclops scheme, metrics, and all three controllers.
`cmd/manager/main.go` and the integration test suite both call it.

2 unit tests added in `pkg/manager/manager_test.go`.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -7,14 +7,11 @@ import (
 	"runtime"
 
 	"github.com/alecthomas/kingpin/v2"
-	"github.com/atlassian-labs/cyclops/pkg/apis"
 	"github.com/atlassian-labs/cyclops/pkg/cloudprovider/builder"
-	"github.com/atlassian-labs/cyclops/pkg/controller/cyclenoderequest"
+	cyclopsmanager "github.com/atlassian-labs/cyclops/pkg/manager"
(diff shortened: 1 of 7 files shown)


hligit (approved):

mwhittington21 (approved):
LGTM

